PR #6, opened 2021-10-04
Add command to update a local Docker Go repo

dagood:
Adds a `dockerupdate` command alongside `dockerupdatepr`.

`dockerupdatepr` requires you to put your branch on a remote for the `-branch` flag, and runs Git operations in a temporary Git directory. This is not very useful when bringing up a whole new major.minor version of Go, because initial work is done locally, and it will likely be manually worked on and submitted as a dev PR.

`dockerupdate` works on an existing Git repo and doesn't run any Git commands, so it's easier to work with when developing nontrivial changes.

Both commands share `UpdateFlags`.
* For `SubmitUpdatePR`, `PRFlags` embeds `UpdateFlags`.
* For `RunUpdate`, `UpdateFlags` is used, plus one directory arg. (It doesn't seem worthwhile to create a command arg struct to contain the single `dockerupdate`-specific arg.)

PR #7, opened 2021-10-04
Handle nil when no build asset JSON was passed

dagood:
The `dockerupdatepr` command accepts `-build-assets-json`, but it's also intended to work without one and simply do an update in place. (Regenerate the manifest JSON and Dockerfiles.) But, `r.buildAssets` will be `nil` if there is no build assets JSON, causing a panic at `"Update "+b.Name+" to "+r.buildAssets.Version`. This PR fixes it by handling `nil`.

The `nil` access only exists in `SubmitUpdatePR`, so it's not necessary for local updates (once PR https://github.com/microsoft/go-infra/pull/6 goes in) but it seems like a reasonable fix for parity between the two.

### DIFF
--- a/buildmodel/commands.go
+++ b/buildmodel/commands.go
@@ -237,7 +237,12 @@ func SubmitUpdatePR(f *PRFlags) error {
 		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


jaredpar (approved):

dagood:
Incorporated into https://github.com/microsoft/go-infra/pull/12

PR #8, opened 2021-10-04
Add simple CI; use gotestsum for AzDO integration

dagood:
Add a simple pipeline that runs tests. Use `gotestsum` to convert the test results into JUnit XML that Azure DevOps can parse and use to show a Test tab. This happens in the dnceng public project.

We only have a few tests, so far: (∅ means no tests exist to run in the package)

```
✓  buildmodel (4ms)
∅  buildmodel/buildassets
∅  buildmodel/dockermanifest
∅  buildmodel/dockerversions
∅  cmd/dockerupdatepr
∅  gitpr

DONE 10 tests in 6.194s
```

Add `CODEOWNERS` to auto-request reviews.

Once this gets merged, I'll add it as a required check for the repo.


dagood:
@microsoft/golang-compiler PTAL, would be good to start adding more tests to the repo, and enable the review auto-request.

chsienki (approved):

qmuntal (changes requested):
`go install` is module-aware since [go1.16](https://golang.org/doc/go1.16#go-command) when using version suffix, such as `go install gotest.tools/gotestsum@v1.7.0` and as it goes through `sum.golang.org` the checksums are also audited.

When using `go install` with a version suffix it ignores local `go.mod` file and does not modifies it, so it safe to be called as a CI step directly from the root module and `enc/ci-tools` is not necessary.

Additionally, since [go1.17](https://golang.org/doc/go1.17#go-command) the `go run` command also accepts version suffixes, so it is possible to run a tool without installing it in a separate command, such as `go install gotest.tools/gotestsum@v1.7.0 --junitfile foo.xml`. The drawback of this approach is that there is no easy way to centralize the version number (other than using bash vars), that why I prefer to install CI tools in their own step.

dagood:
> `go install` is module-aware since [go1.16](https://golang.org/doc/go1.16#go-command) when using version suffix, such as `go install gotest.tools/gotestsum@v1.7.0` and as it goes through `sum.golang.org` the checksums are also audited.

My understanding is that by having the checksums inside our own repo in `eng/ci-tools/go.sum`, we get one extra layer of security. If the gotestsum source and sum.golang.org were compromised (or our connection to them), `eng/ci-tools/go.sum` would prevent installing bad code.

It's just something I noticed when I looked at the tools available--not a pattern I saw anywhere else--so I could be missing some context around `go install` security. 😄 I'm trying to keep a sharp eye on supply chain security but I definitely could have gotten this wrong.

This security likely isn't necessary here (we are only installing/running the tool on a "public" build machine that we already must assume may be compromised) but it seems like good practice to lock down our tools like this when we have an opportunity.

> Additionally, since [go1.17](https://golang.org/doc/go1.17#go-command) the `go run` command also accepts version suffixes, so it is possible to run a tool without installing it in a separate command[...]

Thanks for the tip, and yeah, the tradeoff makes sense.

qmuntal:
This approach adds an additional security layer, you are right. Let's stay on the safe side and keep the go.mod and go.sum files. I can even think on another attack vector: `GOSUMDB` env is unset or set to a malicious server.

Having said this, all the entries of go.mod and go.sum files will be pruned when running `go mod tidy`, and this can happen frequently when upgrading or removing dependencies. The current best practice to track tool dependencies is described here: https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module

It basically says that you need to blank-import your tool dependency in a file called `tools.go` along with the `//go:build tools` build constraint. This way the dependency will be tracked, not removed by `go mod tidy` and at the same time not added in the final binary. Side note: the file name and build constraint name are idiosyncratic, they can have any value.

dagood:
Aha, the `go mod tidy` issue was bugging me, but didn't think to look for a workaround. Thanks, will add that.

eng/ci-tools/tools.go
```diff
@@ -4,9 +4,12 @@
...
+// This file fails to build if 'tools' is set, because gotestsum is a program, not a library. It is
+// necessary so "go install ..." works. There is no reason to build this file, anyway.
+
 //go:build tools
 // +build tools
 
 package ci_tools
 
-import _ "gotest.tools/gotestsum/cmd"
+import _ "gotest.tools/gotestsum"
```
dagood: Interesting story here... with `import _ "gotest.tools/gotestsum"`, ` go build -tags tools ./...` fails:

```
tools.go:12:8: import "gotest.tools/gotestsum" is a program, not an importable package
```

I swapped it to `import _ "gotest.tools/gotestsum/cmd"`, because I thought it was reasonable that this file should be able to build. But that changes the dependencies, so after `go mod tidy`, `go install gotest.tools/gotestsum` fails:

```
C:\Users\dagood\go\pkg\mod\gotest.tools\gotestsum@v1.7.0\cmd\tool\slowest\ast.go:12:2:
  missing go.sum entry for module providing package golang.org/x/tools/go/packages (imported by gotest.tools/gotestsum/cmd/tool/slowest); to add:
        go get gotest.tools/gotestsum/cmd/tool/slowest@v1.7.0
```

So leaving it as an unbuildable `import _ "gotest.tools/gotestsum"` seems correct. There's no reason to build this file, after all. I've added a comment to sum this up.

qmuntal: The errors you are seeing make sense. `gotest.tools/gotestsum` is a `main` package, so `go build`, `go test` and `go vet` will error out when seeing it in an import statement. That's why it is gated by a build constraint that should never be used. On the other hand, `go mod tidy` accepts `main` packages imports and does not take build constraints into account.

Swapping to `import _ "gotest.tools/gotestsum/cmd"` could be a valid move if you want to use any exported API from that package, which we don't 😄. Doing that would also make `gotest.tools/gotestsum` survive `go mod tidy`, as it is the root module of `gotest.tools/gotestsum/cmd`. On the other hand, this import will only add to our `go.sum` the dependencies used by `gotest.tools/gotestsum/cmd`, which does not include `golang.org/x/tools/go/packages` as it is only used in `gotest.tools/gotestsum/cmd/tool/slowest` and this package is not imported by the former.

Hopefully the tools story will be improved once https://github.com/golang/go/issues/48429 is approved and implemented.

eng/pipelines/pr-pipeline.yml
```diff
@@ -0,0 +1,40 @@
...
+      clean: all
+    pool:
+      # Use generic recent LTS.
+      vmImage: ubuntu-20.04
+    steps:
+      - template: steps/checkout-unix-task.yml
+
+      - task: GoTool@0
+        inputs:
+          version: 1.17.1
```
qmuntal, line 22: nit: go 1.17.2 is already out.

Approving anyway!

qmuntal (approved):

eng/pipelines/pr-pipeline.yml
dagood: Good catch--I'll merge https://github.com/microsoft/go-infra/pull/9 first, then change this PR to use the shared `eng/pipelines/steps/init-go.yml` template that specifies 1.17.2.

PR #10, opened 2021-11-04
Create necessary parent dirs when making work dir

dagood:
Also rename func from "GetWorkPathInDir" to "MakeWorkDir" because the func doesn't just get a path, it also creates the directory.

Include a very simple unit test to confirm the parent dir creation behavior.

* Fixes issue with https://github.com/microsoft/go-infra/pull/9
* For https://github.com/microsoft/go/issues/244


dagood:
@microsoft/golang-compiler PTAL, quick fix to finish up the current work on the upstream sync infra.

qmuntal (approved):

PR #11, opened 2021-11-15
Add fips check command

qmuntal:
Adding a tool to check that analyses the Go crypto packages and reports out FIPS compliance data, such as the openssl calls done in each exported API.

`GOROOT` should point to a Go root directory which is checked to a boring-enabled branch, else no boring calls will be identified.

The current output is not machine-friendly, it will be improved if we decide to integrate this tool into CI at some point. 

Example of execution output:

`go run ./cmd/fips crypto/...` 

```sh
...
fips: {crypto/sha1 New true [NewSHA1]}
fips: {crypto/sha1 Sum true [NewSHA1]}
fips: {crypto/sha256 New true [NewSHA256]}
fips: {crypto/sha256 New224 true [NewSHA224]}
fips: {crypto/sha256 Sum256 true [NewSHA256]}
fips: {crypto/sha256 Sum224 true [NewSHA224]}
...
```


cmd/fips/main.go
```diff
@@ -0,0 +1,201 @@
...
+			}
+		}
+		return true
+	})
+	return ret
+}
+
+// searchBoringCalls traverses the node searching for calls to the boring package.
+// Function calls which are not yet resolved will be searched in files.
+func searchBoringCalls(node ast.Node, files []*ast.File, calls []string) []string {
```
dagood, line 157: Nit: I'd call it `findBoringCalls`. `search` implies to me that we're looking through a bunch of boring calls for something specific.

cmd/fips/main.go
```diff
@@ -0,0 +1,201 @@
...
+	ast.Inspect(node, func(n ast.Node) bool {
+		switch n := n.(type) {
+		case *ast.CallExpr: // enableBoring()
+			fn := resolveFun(n, files)
+			if fn != nil {
+				calls = searchBoringCalls(fn.Body, files, calls)
```
dagood, line 163: It seems like the `calls` arg could be removed and the func could just return the slice, so there's no need to pass in `nil` in `processFuncDecl`. Could the func be written like this?

```go
var calls []string
...
calls = append(calls, searchBoringCalls(fn.Body, files))
...
calls = append(calls, n.Sel.Name)
...
return calls
```

cmd/fips/fips_test.go
```diff
@@ -0,0 +1,27 @@
...
+func TestFips(t *testing.T) {
+	cmd := exec.Command("go", "run", ".", "./testdata")
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		t.Logf("fips output:\n%s", output)
+		t.Fatal(err)
+	}
+	got := strings.Split(strings.ReplaceAll(string(output), "\r\n", "\n"), "\n")
+	got = got[:len(got)-1] // remove last empty line jump
+	want := []string{
```
jaredpar, line 19: Is `got / want` a common idiom in Go testing code? Typically used to `actual / expected` in other langs.

cmd/fips/go.mod
```diff
@@ -0,0 +1,11 @@
+module fips
```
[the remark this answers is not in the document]
qmuntal: Good catch. I didn't want to pollute the root `go.mod` with `fips` deps, as we are not using the tool in CI. But I forgot that doing so would skip the tests when running it from an outer module...

cmd/fips/main.go
```diff
@@ -0,0 +1,201 @@
...
+					}
+				}
+			}
+		}
+	}
+}
+
+func parsePackages() ([]*packages.Package, error) {
+	cfg := &packages.Config{
+		Env: append(os.Environ(), "GOOS=linux"),
```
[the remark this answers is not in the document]
qmuntal: This is a tricky one. You cannot set `GOOS=linux` on Windows to check for Linux APIs (and vice-versa) when running the command using `go run`, as it will build the binary for the OS specified in `GOOS`, which in this situation is not to host OS so it won't be executable.

I let it like a hardcoded env variable until I find a better solution, such as passing the GOOS as a cmd argument (open to ideas). But yes, it deserve a comment!

cmd/fips/main.go
```diff
@@ -0,0 +1,201 @@
...
+		for _, src := range pkg.Syntax {
+			for _, decl := range src.Decls {
+				if decl, ok := decl.(*ast.FuncDecl); ok {
+					if !ast.IsExported(decl.Name.Name) {
+						// We are only interested in exported declarations.
+						continue
+					}
+					var report fnReport
+					report.PackageID = pkg.ID
+					report.Name = decl.Name.Name
```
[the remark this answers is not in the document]
qmuntal: Yes it does 👍

cmd/fips/main.go
```diff
@@ -0,0 +1,201 @@
...
+					return false
+				}
+			}
+		}
+		return true
+	})
+	return calls
+}
+
+func processFuncDecl(decl *ast.FuncDecl, files []*ast.File, report *fnReport) bool {
```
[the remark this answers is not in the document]
qmuntal: Changed the code so it is not necessary to return the `bool`.

cmd/fips/main.go
```diff
@@ -0,0 +1,201 @@
...
+	if decl.Body == nil {
+		return false
+	}
+	if decl.Recv != nil {
+		// TODO: check methods.
+		return false
```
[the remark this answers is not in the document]
qmuntal: Done: https://github.com/microsoft/go/issues/278

cmd/fips/main.go
qmuntal: I like it!

cmd/fips/fips_test.go
qmuntal: Yes it is. The tooling that automatically create tabular tests for a given function uses this idiom. Also, searching for `got := ` in the Go source code yields 1265 results in 366 files, and `actual := ` only 32 results in 15 files... probably from someone coming from another lang ;)

cmd/fips/main.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+	seen := make(map[*ast.FuncDecl]struct{})
+	ast.Inspect(decl.Body, func(n ast.Node) bool {
+		switch stmt := n.(type) {
+		// Only search for boring calls that are inside a boring.Enabled if block.
+		case *ast.IfStmt:
+			if hasBoringEnabled(stmt.Cond, seen, files) { // if boring.Enabled() {...}
```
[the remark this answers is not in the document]
qmuntal: I've simplified the `hasBoringEnabled` signature by removing `files`. Now the object resolution takes place before processing an exported function.

I cannot define it as a local function because it is recursive and the go compiler doesn't like it. I could try to refactor the whole algorithm to use methods instead of functions and then add the `seen` map to the parent `struct`, but IMO this will  add more complexity than just passing `seen` as parameter. Thoughts?

I'm not worried about the testability of internal methods, I'm using a black-box testing approach here.

cmd/fips/main.go
```diff
@@ -0,0 +1,234 @@
...
+`
+
+type fnReport struct {
+	PackageID        string
+	Name             string
+	HasBoringEnabled bool
+	BoringCalls      []string
+}
+
+var goos = flag.String("goos", "", "The operating system for which to compile the examinded packaged. Defaults to GOOS")
```
dagood, line 41:
```suggestion
var goos = flag.String("goos", "", "The operating system for which to compile the examined packages. Defaults to GOOS")
```

cmd/fips/main.go
```diff
@@ -0,0 +1,234 @@
...
+Set the appropriate GOROOT when checking std packages not in your default Go toolchain. 
+
+Packages are specified using the notation of "go list".
+This command checks the package in the current directory:
+
+	go run ./cmd/fips
+
+whereas this one checks the std crypto package in GOROOT, as specified:
+
+	go run ./cmd/fips crypto/...
```
dagood, line 31: I think adding an example using `-goos` would be useful, to show why the option exists. 😄

```
The GOOS can also be specified. This allows, for example, examining the openssl APIs while running Windows:

	go run ./cmd/fips crypto/... -goos linux
```

cmd/fips/main.go
```diff
@@ -0,0 +1,234 @@
...
+			packages.NeedName | packages.NeedFiles | packages.NeedCompiledGoFiles |
+			packages.NeedTypes | packages.NeedTypesInfo | packages.NeedTypesSizes,
+	}
+	pkgs, err := packages.Load(cfg, flag.Args()...)
+	if err != nil {
+		return nil, err
+	}
+	return pkgs, nil
+
+}
```
dagood, line 115:
```suggestion
	return pkgs, nil
}
```

cmd/fips/fips_test.go
jaredpar: What tooling are you using to generate the tests?

cmd/fips/main.go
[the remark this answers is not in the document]
dagood: A local function can be recursive, it just needs to have the func variable declared first, to capture:

https://play.golang.org/p/OxHCMxc_S_G
```go
var x func(i int)
x = func(i int) {
	if i > 0 {
		fmt.Printf("%v\n", i)
		x(i - 1)
	}
}
```

But, simplifying works too. 😄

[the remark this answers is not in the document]
dagood: FWIW, though:

> I could try to refactor the whole algorithm to use methods instead of functions [...]

I agree this would be more complexity than it's worth in this case.

Re-reading this bit of the code, now that there are only two args and `seen` is only used in two funcs anyway, I'm personally ok leaving it as you have it now.

cmd/fips/fips_test.go
dagood: I believe VS Code and maybe GoLand use https://github.com/cweill/gotests to generate them. You can see `got`/`want` in some official docs here, too: https://golang.org/doc/code#Testing

qmuntal: VSCode+Go extension provides this capability out of the box. I think https://github.com/cweill/gotests is being used under the hood.

cmd/fips/main.go
qmuntal: done!

cmd/fips/main.go
```diff
@@ -0,0 +1,237 @@
...
+			}
+		}
+	}
+	for _, f := range files {
+		ast.Inspect(f, func(n ast.Node) bool {
+			switch n := n.(type) {
+			case *ast.CallExpr:
+				if ident, ok := n.Fun.(*ast.Ident); ok && ident.Obj == nil {
+					if ident.Obj, ok = funcDecls[ident.Name]; ok {
+						return false
```
dagood, line 139: I'm a little lost trying to find what the function is actually modifying. What I'm expecting to see somewhere is a statement something like `funcDecls[ident.Name].SomethingRef = somethingWeFound`, or some other modification to existing nodes that forces them to get resolved.

Is simply traversing the tree "automatically" causing objects to get resolved, and that's all we need to do here? (Should be noted in func signature if so, IMO.)

dagood (approved):
I'd like clarification on one thing, other than that LGTM. 👍

cmd/fips/fips_test.go
qmuntal: @dagood didn't see your response until I already submitted mine 😄

cmd/fips/main.go
qmuntal: `funcDecls` is just a cache of function declarations, so unresolved indent objects can be resolved later on by just querying the cache using the function name.

The real assignation happens here: `ident.Obj, ok = funcDecls[ident.Name]`. `ident` is a pointer to a node whose function object is undefined, but we have its name, so we can try to resolve it using the declarations in `funcDecls`.

I can see where the confusion come from, so I've added some comments and simplified the code.

dagood (approved):

jaredpar (approved):

qmuntal:
I'm closing this PR as I found that using [https://pkg.go.dev/golang.org/x/tools/go/ssa]https://pkg.go.dev/golang.org/x/tools/go/ssa) instead of `go/ssa` resolved most of the problems I'm having with unresolved symbols.

I'll open a new PR when I have the command ready. Thanks for the reviews anyway, most of them will be useful for the new version as the structure is mostly the same.

PR #12, opened 2021-11-22
Add support for updating go-images dev branches

dagood:
This PR includes a few features/fixes that come together to support updating microsoft/go-images with dev builds from the microsoft/go repository:

* Add a command to locally update a Go Docker images repository. This is useful to set up the branch, e.g. making sure the `manifest.json` gets created correctly and future updates will happen without errors.

* Ignore trailing newlines in the `VERSION` file when generating the build asset JSON in microsoft/go. The official VERSION files never have newlines, so the code had this assumption built-in. The dev branch breaks this by having a trailing newline in its VERSION file, causing a newline to appear in the middle of the version number. (E.g. `1.17.2\n-1`.) This caused problems with the infra. Fix it by using a scanner to read one line.
  * This fix is on the build asset JSON generation side, so we need to update the version of go-infra used by our Go repo.

* Add target branch detection. This lets master and release branch builds continue to get submitted to `microsoft/main`, but let dev branch updates get submitted to their respective dev branches.

* Add a tag prefix feature to the `versions.json` file format in go-images. This lets us distinguish dev branch tags from the upstream-like tags, so they don't get mixed together or overlap.
  * When I was writing this, I didn't like that this is being stuffed into the `versions.json` file. Filed for followup: https://github.com/microsoft/go/issues/281.

---

* For https://github.com/microsoft/go/issues/268


buildmodel/buildassets/buildassets.go
```diff
@@ -148,16 +162,21 @@ func (b BuildResultsDirectoryInfo) CreateSummary() (*BuildAssets, error) {
...
 		if errors.Is(err, os.ErrNotExist) {
 			return defaultVersion
 		}
 		log.Panic(err)
 	}
-	return string(bytes)
+	defer f.Close()
+
+	s := bufio.NewScanner(f)
+	s.Scan()
```
qmuntal, line 178: `bufio.Scan` returns false when the scan stops, either by reaching the end of the input or an error. If you are confident there is no need to handle the error, the idiomatic way to do so is `_ = s.Scan()`.

Nit: Could this function return an error instead of panicking? It is preferable to avoid panics inside non-main packages so they are easy to test and reuse.

buildmodel/buildassets/buildassets_test.go
```diff
@@ -0,0 +1,37 @@
...
+		wantVersion string
+	}{
+		{"Ordinary", "go1.17.3", "go1.17.3"},
+		{"TrailingNewline", "go1.17.3\n", "go1.17.3"},
+		{"MultilineFile", "go1.17.3\nMore information", "go1.17.3"},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			filePath := path.Join(t.TempDir(), "VERSION")
+			err := ioutil.WriteFile(filePath, []byte(tt.text), os.ModePerm)
```
qmuntal, line 26: `io/ioutil` is deprecated since 1.16. Better us [os#WriteFile](https://pkg.go.dev/os#WriteFile) unless this has to be compiled with a previous version.

buildmodel/buildassets/buildassets_test.go
```diff
@@ -0,0 +1,37 @@
...
+			if err != nil {
+				t.Error(err)
```
qmuntal, line 28: `t.Error` is normally used for errors in the expectations of the test. Here we are checking if `ioutil.WriteFile` has done its job as part of the test _setup_, if it fails we cannot run the test, so I would rather us [t.Fatal](https://pkg.go.dev/testing#T.Fatal) instead.

qmuntal (approved):
LGTM with some minor comments.

buildmodel/buildassets/buildassets.go
dagood: `s.Scan` returns a `bool` (whether the scan stopped), not an `error`, so maybe that makes it unnecessary? My IDE warns me about missing error handling, but doesn't warn me about this case. In any case, I'll add it to make it explicit.

I do need to make sure to check `s.Err()` though.

I'll change this to return an error. Some earlier code I wrote had more panics, but I've been moving away from that. (Some of the conversion into `err` I've done so far has been driven by wanting to have the fork sync code finish as many updates as possible, even if an error happens in some particular branch.) Just haven't gotten around to changing this case.

buildmodel/buildassets/buildassets_test.go
jaredpar: Is there a way to trigger diagnostics when using deprecated APIs? Something we can hook into our build to catch it then?

jaredpar (approved):

buildmodel/buildassets/buildassets_test.go
dagood: staticcheck detects deprecated functions (tracking potentially adding that tool with https://github.com/microsoft/go/issues/78), but these `ioutil` functions aren't actually *marked* as deprecated even though https://go.dev/doc/go1.16 did announce it.

https://go.dev/blog/godoc says:

> Sometimes a struct field, function, type, or even a whole package becomes redundant or unnecessary, but must be kept for compatibility with existing programs. To signal that an identifier should not be used, **add a paragraph to its doc comment that begins with “Deprecated:” followed by some information about the deprecation**.

The comments on these deprecated funcs don't follow that guidance:

```
#> go doc ioutil.WriteFile
package ioutil // import "io/ioutil"

func WriteFile(filename string, data []byte, perm fs.FileMode) error
    WriteFile writes data to a file named by filename. If the file does not
    exist, WriteFile creates it with permissions perm (before umask); otherwise
    WriteFile truncates it before writing, without changing permissions.

    As of Go 1.16, this function simply calls os.WriteFile.
```

buildmodel/buildassets/buildassets_test.go
```diff
@@ -0,0 +1,80 @@
...
+	"os"
+	"path"
+	"reflect"
+	"strings"
+	"testing"
+
+	"github.com/microsoft/go-infra/buildmodel/dockerversions"
+)
+
+func TestBuildResultsDirectoryInfo_CreateSummary(t *testing.T) {
```
dagood, line 17: FYI: I created a new test func with a simple baseline to help verify the changes I made to `CreateSummary` when removing `panic`s. It's not meant to be extensive, I just didn't want to continue having absolutely nothing for this code path. 😄

buildmodel/buildassets/buildassets_test.go
dagood: https://go-review.googlesource.com/c/go/+/284777/ has more detail about why these aren't marked:

> I think it's a little soon to call these deprecated.  In general we use "deprecated" markers to tell people that they need to change their code, but this code continues to work just fine.  Let's let people adjust over a couple of releases and then reconsider.

> I agree with Ian that it is too soon to start flagging these uses to people for conversion. People who are still supporting modules built using Go 1.15 _cannot_ start using these functions, so it would be not just useless advice but bad advice. This does not seem to fit "Deprecated" to me. If we do the new inlining-based go fix then these would be good candidates for that. They still wouldn't be deprecated. To me Deprecated implies an important reason not to use the function. There is no important reason not to use these functions, and there is a good reason to keep using them (Go 1.15 compatibility).

qmuntal: Thanks for bringing up the full picture 😄 It is clearly not deprecated, just outdated. I'm not aware of any tool that flags its usage, but I do think it is a good practice to avoid using when possible.

PR #15, opened 2021-12-16
Add sync support for dev branches and tune PR description

dagood:
This PR adds automatic merges from `microsoft/release-branch.go1.17` into `dev/official/go1.17-openssl-fips`.

I touched up the PR description a bit. The main change was to truncate the diff after 200 lines to avoid potential issues with unbounded output. (See code comment.) I went after this right now because with dev branches, the diff could get very large depending on what's being worked on. While I was there, added a couple more things:
* A link to a doc that points at the sync config and entrypoint.
* A collapsible section for the file diff, because usually when you look at one of these PRs, it's because it failed, and this diff isn't useful.

Example PR: https://github.com/dagood/go/pull/9

### DIFF
--- a/cmd/sync/sync.go
+++ b/cmd/sync/sync.go
@@ -56,6 +56,17 @@ var githubPATReviewer = flag.String("github-pat-reviewer", "", "Approve the PR a
 // issues, but other tools have hit some, and it seems reasonable to set a limit ahead of time.
 const maxDiffLinesToDisplay = 200
 
+// GitAuthOption contains a string value given on the command line to indicate what type of auth to
+// use with GitHub URLs.
+type GitAuthOption string
+
+// String values given on the command line. See usage help for details.
+const (
+	GitAuthNone GitAuthOption = "none"
+	GitAuthSSH  GitAuthOption = "ssh"
+	GitAuthPAT  GitAuthOption = "pat"
+)
+
 func main() {
 	var syncConfig = flag.String("c", "eng/sync-config.json", "The sync configuration file to run.")
 	var tempGitDir = flag.String(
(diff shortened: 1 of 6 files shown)


cmd/sync/sync.go
```diff
@@ -126,6 +131,21 @@ func main() {
 	}
 }
 
+// createAuthorizedGitUrl takes a URL, auth options, and returns an authorized URL. The authorized
+// URL may be the same as the original URL, depending on the options given and the URL content.
+func createAuthorizedGitUrl(url string, gitAuthSSH bool, gitAuthPAT bool) string {
```
chsienki, line 136: Nit: it feels weird to be allowed to pass two option flags, but have one of them ignored if the other is true. Does this ever get called with both false? Feels like it might be better to have just a single 'usSSH' and then fall back to PAT if not true?

dagood: I guess *currently* it wouldn't happen because our fork repos are private and need auth, but `false, false` would be for running a sync locally and not actually pushing a PR anywhere. (The Go upstream repos don't require any auth for fetching.) But yeah, this is not a good API--there are only three "real" states for four input values. Should really be an enum or similar.

I'll try changing the bool command line options `-git-auth-ssh` , `-git-auth-pat` to `-git-auth ssh`, `-git-auth pat` to drive this change all the way through.

dagood: This func now takes a `GitAuthOption` (just a typed `string`) and compares against typed consts in a `switch`. Help looks like this:

```
$ go run ./cmd/sync -git-auth wrong
Error: git-auth value 'wrong' is not an accepted value.

[...]
  -git-auth string
        The type of Git auth to inject into upstream and target GitHub URLs for fetch/push access. String options:
         none - Leave GitHub URLs as they are. Git may use HTTPS authentication in this case.
         ssh - Change the GitHub URL to SSH format.
         pat - Add the 'github-user' and 'github-pat' values into the URL.
         (default "none")
  -github-pat string
        Submit the PR with this GitHub PAT, if specified.
[...]
```

chsienki (approved):

qmuntal (approved):

PR #17, opened 2022-01-12
Add doc with reasons for submodules, patch files

dagood:
Currently we're using a fork and in-place modifications. We've always planned to switch to patch files, but I think we should also switch to submodules. So, I put together a few docs about reasons to use patch files and submodules.

We talked about the submodule doc in the team sync today (Jan 12) and think it's ok to start going this way. I added a note after the meeting about `git -c submodule.go.url=https://example.org/go-mirror submodule update`: it's an implementation detail, but it's pretty important for our infra to clone from a mirror for internal builds, and normally submodules *do* have a static URL, so this is a reasonable concern that's worth quickly addressing.

Rendered view: https://github.com/dagood/go-infra/tree/dev/dagood/submodule-doc/docs/fork


jaredpar (approved):

RikkiGibson (commented):
This honestly seems like a great read for anyone considering maintaining a long-term fork of any large project. :)

PR #18, opened 2022-01-14
Add a CI pipeline for tests

leozz37:
Add a CI pipeline for tests using Github Actions


dagood:
Hi! Thanks for the PR. This repository already has CI with an Azure DevOps pipeline, so I don't think it needs a GitHub Actions workflow.

PR #19, opened 2022-01-19
Create submodule sync with mirroring

dagood:
* Add functionality for submodules, mirroring to AzDO, and considering the official/upstream mirror when deciding what to update the submodule to.

* Submodule update commits include the upstream commit message's first line, for context if you don't have quick access to the upstream repo. The message is trimmed if it hits a length limit. I added a test for this func in particular because it's a little intricate, and it's not easy to quickly run manual tests on various upstream messages with full `sync` runs.

* Keep old functionality with a little simplification to fit new scenarios, too. `UpstreamMergeBranches` and `MergeMap` are now just `BranchMap`, with a new `?` token to make `release.branch-go1.17 -> microsoft/?` mappings easy to maintain.
  * This also moves the `master -> microsoft/main` special case into the config file, where it's more visible.

I recommend reviewing with whitespace changes hidden--I wrapped a few existing blocks in `if`s.

Test run: https://dev.azure.com/dnceng/internal/_build/results?buildId=1561104&view=results

* Example submodule update: https://github.com/microsoft/go/pull/377
* Example fork merge: https://github.com/microsoft/go/pull/374
* Example branch-to-branch merge: https://github.com/microsoft/go/pull/375

Example submodule sync config:
```json
  {
    "Upstream": "https://go.googlesource.com/go",
    "UpstreamMirror": "https://github.com/golang/go",
    "Target": "https://github.com/microsoft/go",
    "MirrorTarget": "https://dnceng@dev.azure.com/dnceng/internal/_git/microsoft-go-mirror",
    "BranchMap": {
      "release-branch.go1.17": "dev/dagood/submodule-1.17"
    },
    "SubmoduleTarget": "go"
  }
```

### DIFF
--- a/cmd/sync/model.go
+++ b/cmd/sync/model.go
@@ -8,6 +8,23 @@ package main
 type SyncConfigEntry struct {
 	// Upstream is the upstream Git repository to take updates from.
 	Upstream string
+
+	// UpstreamMirror is an optional upstream-maintained mirror of the Upstream repository.
+	// Specifically, for Upstream 'https://go.googlesource.com/go', the UpstreamMirror is
+	// 'https://github.com/golang/go'.
+	//
+	// When updating a submodule, sync checks both repos and only updates to a version that is in
+	// both. This ensures our dev process doesn't have a strong dependency on one copy of upstream's
+	// code or the other, and makes it reasonable to point the submodule at the GitHub mirror of Go
+	// by default, which has a better appearance in the GitHub UI for a submodule: a clickable link.
+	//
+	// When merging a fork, the check is skipped. A fork inherently mirrors the upstream code, so
+	// there is no reason to look for a common version and hold back.
+	//
+	// If UpstreamMirror is not defined (default), the update simply uses the latest version of
+	// Upstream.
+	UpstreamMirror string
+
 	// Target is the GitHub repository to merge into, then submit the PR onto. It must be an https
 	// github.com URL. Other arguments passed to the sync tool may transform this URL into a
 	// different URL that works with authentication.
@@ -15,15 +32,16 @@ type SyncConfigEntry struct {
 	// Head is the GitHub repository to store the merged branch on. If not specified, defaults to
 	// the value of Target. This can be used to run the PR from a GitHub fork.
 	Head string
+	// MirrorTarget	is an optional Git repository to push the upstream branch to. All mirroring
+	// operations must succeed before sync continues with this sync config entry. The mirror target
+	// is intended to be an internal repo, for reliability and security purposes.
+	MirrorTarget string
 
-	// UpstreamMergeBranches is the list of branches in Upstream to merge into Target, where Target
-	// is a fork repo of Upstream. The branch name in Target associated with each Upstream branch is
-	// determined automatically, including a "microsoft/" prefix to distinguish it.
-	UpstreamMergeBranches []string
-	// MergeMap is a map of source branch name in Upstream to target branch name in Target. This map
-	// should only be used by the configuration file when there is no reasonable way to determine
-	// the target branch name automatically, like UpstreamMergeBranches.
-	MergeMap map[string]string
+	// BranchMap is a map of source branch names in Upstream (keys) to use to update a corresponding
+	// branch in Target (values), where Target is either a fork repo of Upstream or contains a
+	// submodule of Upstream. If the value contains "?" (not a valid branch character), "?" is
+	// replaced with the upstream branch name.
+	BranchMap map[string]string
 
 	// AutoResolveTarget lists files and dirs that Upstream may have modified, but we want to keep
 	// the contents as they are in Target. Normally files that are modified in our fork repos are
@@ -31,4 +49,9 @@ type SyncConfigEntry struct {
 	// some cases this isn't possible. In these cases, Target has in-place modifications that must
 	// be auto-resolved during the sync process.
 	AutoResolveTarget []string
+	// SubmoduleTarget is the path of a submodule in the Target repo to update with the latest
+	// version of Upstream and UpstreamMirror (if specified). If this option is not specified
+	// (default), that indicates the entire Upstream repository should be merged into the Target
+	// repository.
+	SubmoduleTarget string
 }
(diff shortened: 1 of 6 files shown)


cmd/sync/model.go
```diff
@@ -8,27 +8,50 @@ package main
...
+	// code or the other, and makes it reasonable to point the submodule at the GitHub mirror of Go
+	// by default, which has a better appearance in the GitHub UI for a submodule: a clickable link.
+	//
+	// When merging a fork, the check is skipped. A fork inherently mirrors the upstream code, so
+	// there is no reason to look for a common version and hold back.
+	//
+	// If UpstreamMirror is not defined (default), the update simply uses the latest version of
+	// Upstream.
+	UpstreamMirror string
+
```
jaredpar, line 27: Super nit: some fields have blank lines between them, others don't. OCD jared is satsified just writing that down.

cmd/sync/sync.go
```diff
@@ -249,22 +288,88 @@ func syncRepository(dir string, entry SyncConfigEntry) error {
...
 
-		if err := run(newGitCmd("merge", "--no-ff", "--no-commit", b.UpstreamLocalBranch())); err != nil {
-			if exitError, ok := err.(*exec.ExitError); ok {
-				fmt.Printf("---- Merge hit an ExitError: '%v'. A non-zero exit code is expected if there were conflicts. The script will try to resolve them, next.\n", exitError)
-			} else {
-				// Make sure we don't ignore more than we intended.
+		c := changedBranch{Refs: b}
+		var commitMessage string
+
+		if entry.SubmoduleTarget == "" {
```
jaredpar, line 294: Not quite following this part. When there is no submodule target what are we merging here. Is this code flow between branches in the repo?

cmd/sync/model.go
dagood: The idea was they're in groups ("stuff we fetch from", "stuff we push to", "branches", "how is the update carried out"). But a newline snuck in before the `UpstreamMirror` doc comment.

In theory, could use struct embedding to separate these with actual Go syntax (giving us a place to write actual names for these categories)... but that would make it harder to look through the JSON file and correlate each key with one of these fields.

cmd/sync/sync.go
dagood: Yeah, this path handles code flow and merging from an upstream repo into a fork. (`go-images` is still a fork for now. I mentioned earlier that I expected the submodule work to break `go-images`, but it didn't end up being difficult to preserve fork merging, so it will probably still work after the switch.)

cmd/sync/sync.go
```diff
@@ -518,3 +634,17 @@ func combinedOutput(c *exec.Cmd) (string, error) {
...
 }
+
+func createCommitMessageSnippet(message string) (string, error) {
+	s := bufio.NewScanner(strings.NewReader(message))
+	s.Scan()
+	if err := s.Err(); err != nil {
+		return "", err
+	}
+
+	snippet := s.Text()
```
qmuntal, line 645: Is the scanner just advancing one line? If so this can be simplified with something like:

```suggestion
        snippet := message
        if i := strings.IndexByte(snippet, '\n'); i >= 0 {
                snippet = snippet[i+1:]
        }
```

qmuntal (changes requested):

cmd/sync/sync.go
```diff
@@ -48,6 +48,8 @@ var githubUser = flag.String("github-user", "", "Use this github user to submit
 var githubPAT = flag.String("github-pat", "", "Submit the PR with this GitHub PAT, if specified.")
 var githubPATReviewer = flag.String("github-pat-reviewer", "", "Approve the PR and turn on auto-merge with this PAT, if specified. Required, if github-pat specified.")
 
+var azdoDncengPAT = flag.String("azdo-dnceng-pat", "", "Use this Azure DevOps PAT to authenticate to dnceng project HTTPS Git URLs.")
```
qmuntal, line 51: IMO passing secrets via command line is not secure, as they can be leaked in many ways. Could we use an environmental variable instead, such as `AZDO_DNCENG_PAT?

cmd/sync/sync.go
```diff
@@ -249,22 +288,88 @@ func syncRepository(dir string, entry SyncConfigEntry) error {
...
+
+			if len(entry.AutoResolveTarget) > 0 {
+				// Automatically resolve conflicts in specific project doc files. Use '--no-overlay' to make
+				// sure we delete new files in e.g. '.github' that are in upstream but don't exist locally.
+				// '--ours' auto-deletes if upstream modifies a file that we deleted in our branch.
+				if err := run(newGitCmd(append([]string{"checkout", "--no-overlay", "--ours", "HEAD", "--"}, entry.AutoResolveTarget...)...)); err != nil {
+					return err
+				}
+			}
+			c.PRTitle = fmt.Sprintf("Merge upstream `%v` into `%v`", b.UpstreamName, b.Name)
```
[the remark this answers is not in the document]
dagood: Nice, I forgot about that one, and didn't know it had a backticks option. I don't think its handling of backticks is actually safer for the PR title scenario, though--it makes the PR title look like this:

> Merge upstream "main/`strange" into `normal`

vs. the old:

> Merge upstream `main/`strange` into `normal`

I don't find either one particularly better than the other. 😄 

But I do like that it prevents me from accidentally writing some mismatching quotes as I'm coding, so I'll switch over for that.

cmd/sync/sync.go
dagood: I'm concerned about running into `\r\n` rather than just `\n`. *Maybe* not necessary with output from Git commands, but I'm not sure. (I used this approach earlier to read from a file, where `\r\n` handling is more obviously needed.)

But, I'll switch to `strings.IndexAny(snippet, "\r\n")`, which is still simpler than the scanner. (The behavior I'm looking for is `[:i]`, rather than `[i+1:]` though. Which the scanner doesn't exactly make clear! 😄)

cmd/sync/model.go
dagood: For now, removed the empty line above `UpstreamMirror` to make this fit the rough categorization I was going for. Not sure if it's worth doing more to point out the categories, or just make it consistent to avoid questions.

qmuntal (approved):

dagood:
Results look good to me after addressing feedback:
* https://github.com/dagood/go/pull/16
* https://github.com/dagood/go/pull/17

jaredpar (approved):

PR #21, opened 2022-01-26
Add "git-go-patch": patch workflow utility

dagood:
Adds a tool with a few subcommands to help with patch workflows.

I demoed `am` and `fp` in our last sync, and I've added `rebase` to help out with actually editing patches.

I also added a brief readme for the install command and simple workflow guidance. @qmuntal pointed out to me https://go.dev/doc/contribute#git-codereview_install, which has similar setup, so I wrote ours pretty similarly. Readme: https://github.com/dagood/go-infra/blob/dev/dagood/patch-tools/cmd/git-go-patch

I copied some code from microsoft/go for basic patch file application and submodule cleanup--I've separated that out into its own commit so you can review just the diff. The plan is to have the microsoft/go module depend on the code we maintain here in go-infra. (I don't expect this code to be branch-specific.)


patch/patch.go
```diff
@@ -0,0 +1,93 @@
...
+		}
+		if filepath.Ext(entry.Name()) != ".patch" {
+			continue
+		}
+
+		if err := fn(filepath.Join(patchDir, entry.Name())); err != nil {
+			return err
+		}
+	}
+	return nil
```
qmuntal, line 92: This function could be simplified with something like:

```suggestion
        matches, err := filepath.Glob(filepath.Join(rootDir, "patches", "*.patch"))
	if err != nil {
		return err
	}
	for _, match := range matches {
	        if err := fn(match); err != nil {
	        	return err
	        }
	}
	return nil
```

qmuntal (changes requested):

patch/patch.go
dagood: I think I need to sort it, too, then: 
* https://github.com/golang/go/issues/17153

qmuntal:
After a couple of hours of using this tool everything seems to work quite well. It did caught me off guard the `am` command being soo destructive. I had a change already committed an ready to be formatted as a patch, but I used `git go-patch am` instead of `git go-patch fp` and I lost my commit (it was just one line changed 😝).

Would it be possible to detect if the repo is dirty, and if so only apply `am` if a `-force` argument is passed?

dagood:
> I used `git go-patch am` instead of `git go-patch fp` and lost my commit (it was just a one-liner 😝).

Probably not worth it for a one-liner, but you can use `git reflog` inside the submodule to find the old commit hash in the future. 🙂 

I wonder if `am` is actually too similar to `amend`, though. And you could say `fp` is "amending" the patch files. `rebase` is also used to amend commits. I don't think the Git `am` command is *really* popular enough to justify naming our command after it. Same for `git format-patch`, probably.

I think we should call these subcommands `apply` and `extract`. (And keep `rebase`.)

> Would it be possible to detect if the repo is dirty, and if so only apply `am` if a `-force` argument is passed?

It should be, although dirty is a little subjective and IMO it doesn't match what Git thinks:

* After a "fresh" `git go-patch am`, you end up with several commits with essentially random commit hashes on top of the submodule's "default" commit. So, it's already dirty as far as Git status goes. But running `am` again would get you into the exact same state (just with different commit hashes), so it doesn't seem like that should be considered dirty.
* Changes to the base commit. Let's say you ran `submodule update` on `microsoft/main` (no patches involved), and you `pull` a new version of `microsoft/main` that changed the submodule pointer. Git tells you the submodule is dirty. But you can always go back to the old `microsoft/main` commit and run `submodule update` to get the old commit back, so it seems to me like it's ok to update without force.

I think if we make `go-patch am` also record the *post*-patching submodule commit, we can detect the first case. We can already detect the second with the pre-patching record.

(Seeing changed tracked files as dirty is easier--I think `git submodule update` will actually already refuse to run in this case, and we run that first, so our code would already refuse. This can be made more forceful in `-f` mode. 😄)

dagood:
I added a commit for `-f` (and more careful behavior when non-`-f`) earlier, and just added another to use the patching/submodule packages to support the go-images transition to submodules. @qmuntal you might want to re-review, @chsienki @jaredpar I'd appreciate a second review when you have a chance.

buildmodel/commands.go
```diff
@@ -457,8 +461,63 @@ func EnsureDockerfileGenerationPrerequisites() error {
 // any auto-update code runs to detect problems before wasting time on an incompletable update.
 func RunDockerfileGeneration(repoRoot string) error {
 	fmt.Println("Generating Dockerfiles...")
-	var dockerfileUpdateScript = filepath.Join(repoRoot, "eng", "update-dockerfiles.sh")
-	return run(exec.Command("bash", dockerfileUpdateScript))
```
dagood, line 461: For comparison, here's a link to this script that I'm essentially inlining then adding features to in this PR: https://github.com/microsoft/go-images/blob/microsoft/main/eng/update-dockerfiles.sh. (I'm on the fence about linking to the bash script from a code comment--I figure it makes sense not to, since I plan to delete the script in the move to submodules.)

qmuntal (approved):

chsienki (approved):
LGTM

PR #23, opened 2022-02-03
Add Go Docker updates for boring/FIPS branches

dagood:
Add "BranchSuffix" to the Go Docker versions.json format to add "-fips" into the generated tags. Change the update generation code to detect when a build came from a boring/FIPS branch and update the correct versions.json entry with the new build.


qmuntal (approved):
FYI: Go uses https://pkg.go.dev/golang.org/x/mod/semver to parse semantic version strings. Importing it is probably an overkill, but it is a good reference point and its test-suite is also something we would bring.

dagood:
> FYI: Go uses https://pkg.go.dev/golang.org/x/mod/semver to parse semantic version strings. Importing it is probably an overkill, but it is a good reference point and its test-suite is also something we would bring.

Interesting... it seems like it doesn't have a way to get patch version? (Or MajorMinorPatch?) Otherwise I'd go ahead and import it. I poked through the test cases and didn't see anything that seems too useful for our func. (But we aren't trying to parse build parts, or canonicalize.)

chsienki (approved):

PR #24, opened 2022-02-07
build asset JSON: handle `go.*.src.tar.gz` files, improve errors

dagood:
* Fix issue found in testing build for https://github.com/microsoft/go/pull/408
* For https://github.com/microsoft/go/issues/398

Check for `.src.tar.gz` and avoid trying to parse it as a binary build, because it will fail. Also improve the error message in case of failure, and store the source tarball URL in the build asset JSON to make the file more complete.

### DIFF
--- a/buildmodel/buildassets/buildassets.go
+++ b/buildmodel/buildassets/buildassets.go
@@ -55,6 +55,7 @@ func (b BuildAssets) GetDockerRepoTargetBranch() string {
 // less likely to unintentionally break, so some of that information is duplicated here.
 var archiveSuffixes = []string{".tar.gz", ".zip"}
 var checksumSuffix = ".sha256"
+var sourceArchiveSuffix = ".tar.gz"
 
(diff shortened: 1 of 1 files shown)


dagood:
I'm running a test Go internal rolling build with this change vendored in here: https://dev.azure.com/dnceng/internal/_build/results?buildId=1598321&view=results.

dagood:
@microsoft/golang-compiler PTAL
The test Go internal build succeeded, so this should be good to go.

dagood:
I noticed this leaves a blank-ish entry in the build asset JSON because the checksum file of the source tarball isn't handled correctly:

```json
    {
      "env": {
        "GOARCH": "",
        "GOOS": ""
      },
      "sha256": "4f1f09428fcb78786f75eff612e09f9e8271caff03e834f806aa9d967d17544d",
      "url": ""
    },
```

Working on it.

chsienki (approved):

jaredpar (approved):

PR #30, opened 2022-03-24
Move credentials into gitcmd.GitURLAuther

dagood:
The `gitcmd.GitURLAuther` stores credentials, and the `./cmd/sync` code uses it to inject auth data into URLs at the last moment. The short lifetime of the authed URLs makes it easier to trace which variables could be sensitive.

Also fix a small bug with the resulting authed URLs: the old code caused a double slash in `https://github.com//microsoft/go` and an unusual slash in ssh auth mode with `git@github.com:/microsoft/go` (should be `...com:microsoft/go`). These forms actually do happen to work, but they aren't what you'd expect to see and could be confusing.

I initially planned to wrap the Git commands we're using and handle auth with a "client" struct, but this is a more minimal approach we can go with for now. We might still want to wrap the Git commands, for other reasons, and the wrapper can continue to use these authers.

### DIFF
--- a/gitcmd/gitcmd.go
+++ b/gitcmd/gitcmd.go
@@ -1,0 +1,84 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+// Package gitcmd contains utilities for common Git operations in a local repository, including
+// authentication with a remote repository.
+package gitcmd
+
+import (
+	"fmt"
+	"strings"
+)
+
+const githubPrefix = "https://github.com/"
+const azdoDncengPrefix = "https://dnceng@dev.azure.com/"
+
+// GitURLAuther manipulates a Git repository URL (GitHub, AzDO, ...) such that Git commands taking a
+// remote will work with the URL. This is intentionally vague: it could add an access token into the
+// URL, or it could simply make the URL compatible with environmental auth on the machine (SSH).
+type GitURLAuther interface {
+	// InsertAuth inserts authentication into the URL and returns it, or if the auther doesn't
+	// apply, returns the url without any modifications.
+	InsertAuth(url string) string
+}
+
+// GitHubSSHAuther turns an https-style GitHub URL into an SSH-style GitHub URL.
+type GitHubSSHAuther struct{}
+
+func (a GitHubSSHAuther) InsertAuth(url string) string {
+	if strings.HasPrefix(url, githubPrefix) {
+		return fmt.Sprintf("git@github.com:%v", strings.TrimPrefix(url, githubPrefix))
+	}
+	return url
+}
+
+// GitHubPATAuther adds a username and password into the https-style GitHub URL.
+type GitHubPATAuther struct {
+	User, PAT string
+}
+
+func (a GitHubPATAuther) InsertAuth(url string) string {
+	if a.User != "" && a.PAT != "" && strings.HasPrefix(url, githubPrefix) {
+		return fmt.Sprintf(
+			"https://%v:%v@github.com/%v",
+			a.User, a.PAT, strings.TrimPrefix(url, githubPrefix))
+	}
+	return url
+}
+
+// AzDOPATAuther adds a PAT into the https-style Azure DevOps repository URL.
+type AzDOPATAuther struct {
+	PAT string
+}
+
+func (a AzDOPATAuther) InsertAuth(url string) string {
+	if a.PAT != "" && strings.HasPrefix(url, azdoDncengPrefix) {
+		url = fmt.Sprintf(
+			// Username doesn't matter. PAT is identity.
+			"https://arbitraryusername:%v@dev.azure.com%v",
+			a.PAT, strings.TrimPrefix(url, azdoDncengPrefix))
+	}
+	return url
+}
+
+// NoAuther does nothing to URLs.
+type NoAuther struct{}
+
+func (a NoAuther) InsertAuth(url string) string {
+	return url
+}
+
+// MultiAuther tries multiple authers in sequence. Stops and returns the result when any auther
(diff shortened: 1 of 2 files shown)


gitcmd/gitcmd.go
```diff
@@ -0,0 +1,84 @@
...
+	InsertAuth(url string) string
+}
+
+// GitHubSSHAuther turns an https-style GitHub URL into an SSH-style GitHub URL.
+type GitHubSSHAuther struct{}
+
+func (a GitHubSSHAuther) InsertAuth(url string) string {
+	if strings.HasPrefix(url, githubPrefix) {
+		return fmt.Sprintf("git@github.com:%v", strings.TrimPrefix(url, githubPrefix))
+	}
```
qmuntal, line 31: I'm not sure if this module can contain go1.18 features, but if so you can use my favorite addition to the standard library in this release, [strings.Cut](https://pkg.go.dev/strings#Cut):

```suggestion
        if _, after, found := strings.Cut(url, githubPrefix); found {
        	return fmt.Sprintf("git@github.com:%v", after)
        }
```

This apply to other HasPrefix+TrimPrefix occurrences in this file.

gitcmd/gitcmd.go
```diff
@@ -0,0 +1,84 @@
...
+			"https://arbitraryusername:%v@dev.azure.com%v",
+			a.PAT, strings.TrimPrefix(url, azdoDncengPrefix))
+	}
+	return url
+}
+
+// NoAuther does nothing to URLs.
+type NoAuther struct{}
+
+func (a NoAuther) InsertAuth(url string) string {
```
qmuntal, line 67: nit: receiver can be omitted.

```suggestion
func (NoAuther) InsertAuth(url string) string {
```

gitcmd/gitcmd.go
```diff
@@ -0,0 +1,84 @@
...
+type GitURLAuther interface {
+	// InsertAuth inserts authentication into the URL and returns it, or if the auther doesn't
+	// apply, returns the url without any modifications.
+	InsertAuth(url string) string
+}
+
+// GitHubSSHAuther turns an https-style GitHub URL into an SSH-style GitHub URL.
+type GitHubSSHAuther struct{}
+
+func (a GitHubSSHAuther) InsertAuth(url string) string {
```
qmuntal, line 28: nit: receiver can be omitted.

```suggestion
func (GitHubSSHAuther) InsertAuth(url string) string {
```

gitcmd/gitcmd.go
```diff
@@ -0,0 +1,84 @@
...
+	"strings"
+)
+
+const githubPrefix = "https://github.com/"
+const azdoDncengPrefix = "https://dnceng@dev.azure.com/"
+
+// GitURLAuther manipulates a Git repository URL (GitHub, AzDO, ...) such that Git commands taking a
+// remote will work with the URL. This is intentionally vague: it could add an access token into the
+// URL, or it could simply make the URL compatible with environmental auth on the machine (SSH).
+type GitURLAuther interface {
```
qmuntal, line 19: `gitcmd.URLAuther` is shorted and gives the same info as `gitcmd.GitURLAuther`.

`GIT` prefix seems redundant. The package name already starts with `git` and Go naming conventions strongly recommend not repeating scopes in symbols.

qmuntal (changes requested):

gitcmd/gitcmd.go
dagood: Hmm, go-infra is used in microsoft/go by https://github.com/microsoft/go/blob/microsoft/main/eng/_util/go.mod (each release branch). But `_util` won't use this particular package, and even if it did (or will), it's fine for `_util` to require a new version of Go. It's fine to use go1.18 to run the infra/wrappers that build go1.17.x.

This is indeed a really great new func--using it. 😄

dagood: Noticed that `Cut` isn't quite `CutPrefix`, and to really be equivalent, need to check `before` too:

```go
	if before, after, found := strings.Cut(url, githubPrefix); before == "" && found {
```

This is pretty unlikely to happen for a significant chunk of a URL like this, but this made me think... I could add a really simple `cutPrefix` func to the package based on the `strings.Cut` API design, and implement it directly to avoid even the possibility of issues with standard library versioning:

```go
func cutPrefix(s, prefix string) (after string, found bool) {
	if strings.HasPrefix(s, prefix) {
		return s[len(prefix):], true
	}
	return s, false
}
```

qmuntal (approved):

gitcmd/gitcmd.go
```diff
@@ -0,0 +1,95 @@
...
+// remote will work with the URL. This is intentionally vague: it could add an access token into the
+// URL, or it could simply make the URL compatible with environmental auth on the machine (SSH).
+type URLAuther interface {
+	// InsertAuth inserts authentication into the URL and returns it, or if the auther doesn't
+	// apply, returns the url without any modifications.
+	InsertAuth(url string) string
+}
+
+// GitHubSSHAuther turns an https-style GitHub URL into an SSH-style GitHub URL.
+type GitHubSSHAuther struct{}
```
jaredpar, line 26: Given that we removed the `Git` prefix from `GitURLAuther` I would have expected that it is removed here too. Is there a reason it was kept?

dagood: Well, here it isn't a Git prefix, rather a GitHub prefix. `HubSSHAuther` is confusing, and it is GitHub specific (other services don't define SSH URLs the same way, if they have them at all).

jaredpar: makes sense

jaredpar (approved):

PR #35, opened 2022-03-30
Add "git go-patch extract" patch renumbering

dagood:
* https://github.com/microsoft/go/issues/505

Example usage in a patch file (leading `0` optional, just matches 4-digit convention):

```
Subject: [PATCH] Add vendored go-crypto-openssl module

github.com/microsoft/go-infra/cmd/git-go-patch command: patch number 0100
---
```

That prefix hopefully makes it at least a little traceable if the message ends up somewhere we don't expect. I made it so it'll be easy to add another command with that prefix later, if any reasons for it come up.

I think it would make sense for `git go-patch apply` to get a flag that makes it strip out these lines (for submission to upstream). But, we don't need that for the FIPS patches, so filing an issue for that as followup:

* https://github.com/microsoft/go/issues/512

### DIFF
--- a/cmd/git-go-patch/extract.go
+++ b/cmd/git-go-patch/extract.go
@@ -4,13 +4,20 @@
 package main
 
 import (
+	"bufio"
+	"errors"
 	"flag"
+	"fmt"
+	"log"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
+	"strings"
 
 	"github.com/microsoft/go-infra/executil"
 	"github.com/microsoft/go-infra/patch"
+	"github.com/microsoft/go-infra/stringutil"
 )
 
 const extractSummary = "Format each new commit in the submodule as a patch file."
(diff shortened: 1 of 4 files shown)


cmd/git-go-patch/extract.go
```diff
@@ -70,15 +101,142 @@ var extract = subcommand{
...
+		}
+	}
+	return cmds, nil
+}
+
+func renameUnlessExists(path, newPath string) error {
+	_, err := os.Stat(newPath)
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return os.Rename(path, newPath)
```
dagood, line 221: Moving over to Linux where my temp is set up differently, I see this only works on one partition. (And maybe more limitations--the doc notes that it's only really intended to work within a single directory.) This isn't general-purpose like `mv`. Changing the code to use copies.

qmuntal (approved):

PR #38, opened 2022-04-15
Handle "armv6l" in build asset JSON and manifest.json

dagood:
* For https://github.com/microsoft/go/issues/519
* Fixes a problem in https://github.com/microsoft/go/pull/523

I had to make updates pretty much throughout the microsoft/go -> microsoft/go-images update process to account for armv6l builds. (AKA `GOARCH=arm GOARM=6`, `arm/v6`, and `arm32v6`.)

The arm64 path needed a fix too. It's called arm64v8 in Docker Hub and .NET Docker. This PR changes our tagging structure from `arm64` to call it `arm64v8`, e.g. in `oss/go/microsoft/golang:1.18.1-1-bullseye-arm64v8`. This makes our naming a little closer to https://hub.docker.com/r/arm64v8/golang/. (However, most people will probably get this image using our manifest tags, which don't include the `arm64v8` string anyway.)

This PR adds broad tests for the update funcs, which helped a lot to make sure this doesn't regress the existing architectures. They compare the entire versions.json and manifest.json after the update call, so I found the https://github.com/go-test/deep module to help keep the results readable. Test failure logs look like this:

```
=== RUN   TestUpdateManifest
    buildmodel_test.go:158: Repos.slice[0].Images.slice[0].Platforms.slice[2].Tags.map[1.18.1-1-bullseye-arm32v7]: {} != <does not have key>
    buildmodel_test.go:158: Repos.slice[0].Images.slice[0].Platforms.slice[2].Tags.map[1.18.1-1-bullseye-arm32v7-examplefail]: <does not have key> != {}
--- FAIL: TestUpdateManifest (0.01s)
```

I also had the tests write the actual result to a file if they don't match, so it's easy to copy the actual result over the wanted result file when something intentionally changes.

I tried out a go-images build with an updated manifest here: https://dev.azure.com/dnceng/internal/_build/results?buildId=1720391&view=results. (This kind of dev build doesn't actually produce manifests, so if you try it out, make sure to use the full tag with arch specified.)


buildmodel/testdata/.gitignore
```diff
@@ -0,0 +1,4 @@
+# This test data directory contains input/output JSON files and the wanted output. Tests also
+# generate a "*-actual.json" file when they run to make it easy for a dev to rewrite the wanted file
+# when the desired behavior changes or new cases are added. Ignore this artifact in Git.
+*-actual.json
```
jaredpar, line 4: Is it common for Go testing to generate files into the `testdata` directory like this? Generally I think of using `%TEMP%` or the like for test code I/O to keep the source tree pure. In the past teams like VS would even go so far as to make the source tree directory completley read only to enforce this

buildmodel/dockerversions/dockerversions.go
```diff
@@ -85,13 +85,24 @@ func (a ArchEnv) GoImageArchSuffix() string {
...
 	if a.GOOS != "linux" {
 		s = a.GOOS + "-" + s
 	}
 	return s
 }
+
+// GoImageArchKey replicates upstream behavior, creating the arch string used by
+// docker-library/golang and .NET Docker to identify this arch in the versions.json and
+// manifest.json files.
+func (a ArchEnv) GoImageArchKey() string {
```
jaredpar, line 102: Curious: is there a reason you're passing by value instead of pointer here? It's now a triple word struct which is my general line for "maybe start passing via ref".

jaredpar (approved):

buildmodel/testdata/.gitignore
dagood: It seems pretty common. I caught a glance of this idea a while back in `gotestsum`'s readme, `-update-golden`: https://github.com/gotestyourself/gotestsum#re-running-failed-tests. I took a fresh look now, and saw a couple articles:

* https://ieftimov.com/posts/testing-in-go-golden-files/
* https://medium.com/soon-london/testing-with-golden-files-in-go-7fccc71c43d3

That tipped me off that this pattern is actually used in golang/go: https://github.com/golang/go/blob/91b9915d3f6f8cd2e9e9fda63f67772803adfa03/src/cmd/gofmt/gofmt_test.go#L18

**But** this is a little different from what I have here--going further. Passing `-update` to the test command is definitely more intentional than writing a new file upon failure. Much cleaner, and perhaps orthogonal to making the test dir readonly.

I'll change it to an `-update` flag, I like that workflow better anyway. 😄

buildmodel/dockerversions/dockerversions.go
dagood: It looks like three string fields -> three sort-of-slices ([pointer + length, no cap](https://github.com/golang/go/blob/91b9915d3f6f8cd2e9e9fda63f67772803adfa03/src/runtime/string.go#L238-L241)) -> 6 words.

(Worked backwards from `unsafe.Sizeof` and noticed that `[]rune` is 3 words but `string` is 2. 🙂)

No reason to keep it as a value receiver here--changing.

dagood: FWIW, these guidelines look good to me: https://github.com/golang/go/wiki/CodeReviewComments#receiver-type. There are *very* few reasons there to use a value receiver: if the receiver type is a map/func/chan, or if it's tiny and demonstrably helps perf.

buildmodel/buildmodel_test.go
```diff
@@ -5,14 +5,24 @@ package buildmodel
...
 
+	"github.com/go-test/deep"
 	"github.com/microsoft/go-infra/buildmodel/buildassets"
 	"github.com/microsoft/go-infra/buildmodel/dockermanifest"
 	"github.com/microsoft/go-infra/buildmodel/dockerversions"
 )
 
+var update = flag.Bool("update", false, "Update the golden files instead of failing.")
+
+func TestMain(m *testing.M) {
```
qmuntal, line 21: You don't need to call `flag.Parse()` in test code, it's done implicitly, and it's only necessary if `TestMain` depends on a flag, which is not the case:
https://pkg.go.dev/testing#hdr-Main

buildmodel/buildassets/buildassets_test.go
```diff
@@ -44,8 +63,10 @@ func TestBuildResultsDirectoryInfo_CreateSummary(t *testing.T) {
 		t.Errorf("CreateSummary() error is not wanted: %v", err)
 		return
 	}
-	if !reflect.DeepEqual(got, want) {
-		t.Errorf("CreateSummary() got = %v, want %v", got, want)
+	if diff := deep.Equal(got, want); diff != nil {
+		for _, d := range diff {
```
qmuntal, line 67: The loop is not necessary to pretty print `diff`. You can pass it directly to `t.Error`.

Unrelated: I'm a big fan of `go-tets/deep` 😄

qmuntal (changes requested):

buildmodel/buildassets/buildassets_test.go
qmuntal: Convinced me 😸

PR #40, opened 2022-04-27
Add GitHub tagging and release creation automation using GitHub and Azure Pipelines APIs

dagood:
I recommend reviewing with "hide whitespace" enabled. I changed the subcommand implementation to use an interface rather than have a func in a struct, so a lot of code is unindented by one level.

---

* For https://github.com/microsoft/go/issues/423
* Also implements https://github.com/microsoft/go/issues/463

Finish off post-build release process automation for microsoft/go. This gets rid of a lot of copy-pasting version numbers so the process is less tedious and error-prone.

Here's an example of the comments that the pipeline (optionally) leaves on the release issue: https://github.com/dagood/go-release-test/issues/3.
Example release/tag: https://github.com/dagood/go-release-test/releases/tag/v1.18.1-1

Pipeline run UI looks like this:

image

I plan to have the variable group pass through all the release steps to make it reasonable to run a fake release without worrying too much about messing up the main repos.

Individual steps are toggleable so if a build partially fails, you can run a retry that only includes the steps that still need to happen. If the steps were each idempotent, these toggles wouldn't be needed, but I didn't want to try to rig up reliable idempotency on top of the GitHub APIs--at least not yet. I think keeping it simple will work out fine.

This adds a few libraries:
* https://github.com/google/go-github - API to create the tag, GitHub release, and comment on the issue.
  * https://golang.org/x/oauth2 - For go-github auth.
* https://github.com/microsoft/azure-devops-go-api - Get more details about the build being released, like commit to tag.

I didn't switch the existing code over to use go-github yet. Note: go-github uses the GitHub v3 API, so it actually can't do auto-merges, which requires v4. But the GitHub v4 API is incomplete (it [doesn't support releases](https://github.community/t/creating-a-github-release-using-github-api-v4/13633/6)), so we can't simply use a newer library that supports v4.

---

`github.com/microsoft/go-infra/subcmd`

This package makes it easier to implement subcommands. I moved the subcommand functionality from git-go-patch over here and changed it to use interfaces, which are more typical for this kind of sharing.

I changed the existing `update-aka-ms` command into the `releasego akams` subcommand, which meant some tweaks like adding the ability to handle arbitrary args by implementing `OptionArgTaker` instead of only `Option`.

I wanted to use subcommands for `releasego` to give the commands a clear namespace (so subcommand names don't have to be as descriptive and can trivially share code) and so we can build a single app that gets used throughout the steps of the pipeline for simplicity.


qmuntal (commented):
Did an initial review and everything looks great!

I didn't find anything to comment, so I will let it sink in and tomorrow I'll give it another shot 😄

cmd/git-go-patch/apply.go
```diff
@@ -14,83 +14,90 @@ import (
...
+	if err != nil {
+		return err
+	}
 
-		if !*noRefresh {
-			if err := submodule.Reset(rootDir, *force); err != nil {
-				return err
-			}
-		}
+	goDir := filepath.Join(rootDir, "go")
```
jaredpar, line 59: The `extract, `apply` and `rebase` commands have the following pattern: 

```go
        rootDir, err := findOuterRepoRoot()
	if err != nil {
		return err
	}

	goDir := filepath.Join(rootDir, "go")
 ``` 

Is it worth factoring that out?

File: cmd/git-go-patch/apply.go [hunk 2]
```diff
@@ -14,83 +14,90 @@ import (
...
-		if err != nil {
-			return err
-		}
+	// Record the pre-patch commit. We must do this before applying the patch: if patching
+	// fails, the user needs to be able to fix up the patches inside the submodule and then run
+	// "git go-patch extract" to apply the fixes to the patch files. "extract" depends on the
+	// pre-patch status file. Start by ensuring the dir exists, then write the file.
+	if err := os.MkdirAll(getStatusFileDir(rootDir), os.ModePerm); err != nil {
+		return err
+	}
```
jaredpar, line 85:
```suggestion
	}
	
```

Nit: consistency

cmd/git-go-patch/extract.go
```diff
@@ -43,145 +56,138 @@ Each command starts with "` + commandPrefix + `", then:
   distinct groups of patches. Making the first patch in each group start at a consistent number can
   help to avoid unnecessary filename conflicts when porting changes between branches.
 ` + repoRootSearchDescription
+}
+func (e extractCmd) Handle(p subcmd.ParseFunc) error {
```
jaredpar, line 60: One thought I had reading through this is there is a lot of IO happening and when it fails we resort to 

```go
return err
```

I'm considering how that will look when we have an error in an AzDO run. It will result in a failure that has an IO message in the logs but it seems like it will be really hard to track that back down to what failed. Basically if we have a "Access Denied" error in the logs how will we tell which of the various IO paths in this code was responsible? 

Is it worth adding logs on the failure cases so we have a trail to look at?

cmd/releasego/get-asset-version.go
```diff
@@ -0,0 +1,87 @@
...
+	buildAssetJSON := flag.String("build-asset-json", "", "[Required] The path of a build asset JSON file to read.")
+
+	validateVersionFlag := flag.String(
+		"version", "",
+		"A Microsoft-built Go version, in 1.2.3-1[-fips] format.\n"+
+			"If specified, must match the build asset JSON's version or this command will fail.\n"+
+			"The string 'nil' is treated as the same as not setting the value, for use in CI where empty string can't be used.\n"+
+			"Optionally use this to validate expectations.")
+
+	setVariable := flag.String("set", "", "An AzDO variable name to set.")
```
chsienki, line 53: nit: can we call this setVariable or something that makes it more obvious what it does from logs when you don't have the help?

cmd/releasego/repo-release.go
```diff
@@ -0,0 +1,175 @@
...
+
+			_, _, err = client.Repositories.UploadReleaseAsset(
+				ctx,
+				owner, name,
+				*release.ID,
+				&github.UploadOptions{Name: filename},
+				file)
+			return err
+		}); err != nil {
+			return fmt.Errorf("failed to upload %#q to release %v: %w", p, *release.ID, err)
```
chsienki, line 128: If we fail to upload a file, we should fail the release. Does this happen here? Possibly it does, just my lack of go knowledge showing :)

chsienki: Ah, we're outside of the `retry` so this will return from `Handle` which will kick in the defered cleanup

subcmd/subcmd.go
```diff
@@ -0,0 +1,96 @@
...
+	"fmt"
+	"os"
+)
+
+// ParseFunc parses all flags that have been set up. Include extra handling for "-h" help flag
+// detailed output, invalid args, and error conditions.
+type ParseFunc func() error
+
+// Option implements one subcommand option that the user may pick.
+type Option interface {
```
qmuntal, line 18: As discussed offline, is not modeling behavior but just enumerating properties. The only behavior here are `Handle` and `ArgsSummary`.

Therefore, it would be better to define it as a struct which contains good old props and two functions pointers.

qmuntal (changes requested):

cmd/releasego/get-asset-version.go
dagood: Sure, going one step further and changing it to `-set-azdo-variable`. (It's not like length is really a concern here, since it's pretty useless outside of CI. 😄)

cmd/releasego/repo-release.go
dagood: Yep, `err := retry(...)` gets the last inner error *if* it exceeds the retry count, then `if err := retry(...); err != nil {` continues here, and this return is in the function's scope.

cmd/git-go-patch/extract.go
dagood: In a lot of cases I think we can pretty easily figure out where an error happened by looking at the last positive/progress log line. But, I try to always remember to wrap errors when I think there might be some ambiguity, like `return fmt.Errorf("unable to chop the carrot: %w", err)` when `err` could be `knife is dull` but the func is making an entire bowl of soup.

In my experience, IO errors always mention the filename, so there are a decent number of situations where even without wrapping, the error would still be obvious.

This seems pretty subjective to me, and wrapping too much is definitely better than not wrapping enough, so I'd be happy to put more in. (Just maybe not in `git-go-patch` because this isn't new code, only unindented.)

qmuntal (approved):

PR #42, opened 2022-05-18
Add microsoft/go release-build automation

dagood:
* For https://github.com/microsoft/go/issues/423

There's more work to do in the future for this pipeline:
* Docker images
* https://github.com/microsoft/go/issues/470
* https://github.com/microsoft/go/issues/542
* https://github.com/microsoft/go/issues/575

---

Add almost all of the release-build section of https://github.com/microsoft/go-infra/blob/main/docs/release-process/README.md#2-release-build. (This PR updates the doc with the specifics.) 

Adds a handful of new tools to `releasego`:

* `build-pipeline` - Run a build of an AzDO pipeline. Used to trigger a microsoft/go build and other release steps.
  * Doesn't wait--see `wait-build`.
* `get-merged-pr-commit` - Get the commit that resulted from merging the given GitHub PR.
  * If the PR hasn't been merged, wait. (Poll.) If something other than merging happens, fail.
  * It also checks if the PR is still will definitely not merge without human intervention and fail immediately to let a human know. This way it doesn't wait around forever for something that will never happen.
*  `get-target-branch` - Gets the target microsoft/go branch for the targeted version number.
* `sync` - Runs the ordinary upstream-sync process and submits a PR, but specialized for the release process.
  * See command desc for more info.
  * I pulled out a lot of old `cmd/sync` functionality into a `sync` package to share between `cmd/sync` and `cmd/releasego/sync.go`. I changed `gitpr.FindExistingPR` to return more data about the PR so that the release pipeline can run further steps on it.
* `wait-azdo-commit` - Wait for a commit (the result of merging the `sync` PR) to show up in the internal mirror of the GitHub repo.
* `wait-build` - Wait for an AzDO build to complete.

`eng/pipelines/release-build-pipeline.yml` combines these to run the release steps.

[Link to the `go-release-config` variable group.](https://dev.azure.com/dnceng/internal/_library?itemType=VariableGroups&view=VariableGroupView&variableGroupId=156&path=go-release-config)

This PR uses the dependencies I added to the `go.mod` in my last PR:

https://github.com/microsoft/go-infra/blob/850c558938dd65b867bf76b1cb44027ad5606bf8/go.mod#L10-L12

---

Example successful run from the top: https://dev.azure.com/dnceng/internal/_build/results?buildId=1780270&view=results
Uses a temp testing repo on GitHub where I manually merged the sync PR and pushed it to a dev branch in AzDO to keep the run going.

Example run that (intentionally) ran out of time waiting for the AzDO mirror to sync: https://dev.azure.com/dnceng/internal/_build/results?buildId=1780032&view=results.

### DIFF
--- a/azdo/azdo.go
+++ b/azdo/azdo.go
@@ -1,0 +1,78 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package azdo
+
+import (
+	"errors"
+	"flag"
+	"fmt"
+
+	"github.com/microsoft/azure-devops-go-api/azuredevops"
+	"github.com/microsoft/azure-devops-go-api/azuredevops/build"
+)
+
+// ClientFlags is a set of command-line flags always used in go-infra to access the AzDO APIs.
+type ClientFlags struct {
+	// Org is the base URL of the organization, such as 'https://dev.azure.com/dnceng/'
+	Org *string
+	// Proj is the project, such as 'public' or 'internal'.
+	Proj *string
+	// PAT is the Personal Access Token to use.
+	PAT *string
+}
+
+// BindClientFlags creates a ClientFlags struct where each field is set up as a flag.
+func BindClientFlags() *ClientFlags {
+	return &ClientFlags{
+		Org:  flag.String("org", "", "[Required] The AzDO organization URL."),
+		Proj: flag.String("proj", "", "[Required] The AzDO project URL."),
+		PAT:  flag.String("azdopat", "", "[Required] The Azure DevOps PAT to use."),
+	}
+}
+
+// EnsureAssigned can be called after "flag.Parse()" to ensure all required flags were specified.
+func (c *ClientFlags) EnsureAssigned() error {
+	if *c.Org == "" {
+		return errors.New("no AzDO org specified")
+	}
+	if *c.Proj == "" {
+		return errors.New("no AzDO project URL specified")
+	}
+	if *c.PAT == "" {
+		return errors.New("no AzDO PAT specified")
+	}
+	return nil
+}
+
+// NewConnection creates an AzDO connection based on the given flags.
+func (c *ClientFlags) NewConnection() *azuredevops.Connection {
+	return azuredevops.NewPatConnection(*c.Org, *c.PAT)
+}
+
+// GetBuildWebURL finds the web/UI URL (not API endpoint URL) in the given AzDO Build, if it exists.
+func GetBuildWebURL(b *build.Build) (string, bool) {
+	links, ok := b.Links.(map[string]interface{})
+	if !ok {
+		return "", false
+	}
+	web, ok := links["web"].(map[string]interface{})
+	if !ok {
+		return "", false
+	}
+	href, ok := web["href"]
+	if !ok {
+		return "", false
+	}
+	s, ok := href.(string)
+	if !ok {
+		return "", false
+	}
+	return s, true
+}
+
+// SetPipelineVariable uses an AzDO logging command to set a variable in the pipeline.
+// https://github.com/Microsoft/azure-pipelines-tasks/blob/master/docs/authoring/commands.md
+func SetPipelineVariable(name, value string) {
+	fmt.Printf("##vso[task.setvariable variable=%v]%v\n", name, value)
+}
--- a/cmd/releasego/build-pipeline.go
+++ b/cmd/releasego/build-pipeline.go
@@ -129,6 +129,7 @@ func handleBuildPipeline(p subcmd.ParseFunc) error {
 	if err != nil {
 		return err
 	}
+        defer resp.Body.Close()
 	bodyBytes, err := io.ReadAll(resp.Body)
 	if err != nil {
 		return err
--- a/cmd/releasego/build-pipeline.go
+++ b/cmd/releasego/build-pipeline.go
@@ -4,6 +4,7 @@
 package main
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"flag"
@@ -12,7 +13,6 @@ import (
 	"log"
 	"net/http"
 	"strconv"
-	"strings"
 
 	"github.com/microsoft/azure-devops-go-api/azuredevops/build"
 	"github.com/microsoft/go-infra/azdo"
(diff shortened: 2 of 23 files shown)


dagood:
(I rebased to split out [Centralize sync package and add "releasego sync"](https://github.com/microsoft/go-infra/pull/42/commits/43b35a15806e1d766629219ef63395f205ff671c) so it's easier to review the sync changes in context with each other.)

jaredpar:
Is it best to review this commit by commit or just look at the final result?

dagood:
Hmm. I'll squash this to two commits: the sync changes and everything else. I think reviewing those two diffs individually is the way to go.

cmd/releasego/build-pipeline.go
```diff
@@ -0,0 +1,157 @@
...
+		// AzDO UI--this is unrelated to the template parameters.
+		"parameters": string(variablesJSON),
+	}
+
+	bodyJSON, err := json.MarshalIndent(body, "", "  ")
+	if err != nil {
+		return err
+	}
+
+	req, err := http.NewRequestWithContext(ctx, "POST", url, strings.NewReader(string(bodyJSON)))
```
qmuntal, line 116: You can use `bytes.NewReader` and avoid the cast.

cmd/releasego/build-pipeline.go
```diff
@@ -0,0 +1,157 @@
...
+	// Based on https://github.com/microsoft/azure-devops-go-api/blob/00dac5c867394a3c5ca4e12b6965d7625a1588c6/azuredevops/client.go#L172-L181
+	req.Header.Add("Authorization", azdoFlags.NewConnection().AuthorizationString)
+	req.Header.Add("Accept", "application/json;api-version=7.1-preview.7")
+	req.Header.Add("Content-Type", "application/json;charset=utf-8")
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return err
+	}
+	bodyBytes, err := io.ReadAll(resp.Body)
```
qmuntal, line 132: Remember to close the Body.

```suggestion
        defer resp.Body.Close()
	bodyBytes, err := io.ReadAll(resp.Body)
```

qmuntal (changes requested):
Looks really good, just some small comments.

dagood:
I wrote a doc about how to use this pipeline--might help with context for reviewing:

* https://github.com/microsoft/go-infra/pull/43

dagood:
The pipeline should be usable for normal release branch releases from upstream, but there are a few things I need to implement in `releasego sync` to make it usable for boring branch and microsoft-revision releases:

* https://github.com/microsoft/go/issues/470
* https://github.com/microsoft/go/issues/575

I'm working on those, but I think this PR is reviewable/mergeable as-is to make incremental progress.

azdo/azdo.go
```diff
@@ -0,0 +1,78 @@
...
+	// PAT is the Personal Access Token to use.
+	PAT *string
+}
+
+// BindClientFlags creates a ClientFlags struct where each field is set up as a flag.
+func BindClientFlags() *ClientFlags {
+	return &ClientFlags{
+		Org:  flag.String("org", "", "[Required] The AzDO organization URL."),
+		Proj: flag.String("proj", "", "[Required] The AzDO project URL."),
+		PAT:  flag.String("azdopat", "", "[Required] The Azure DevOps PAT to use."),
```
jaredpar, line 30: I'm missing how these flags end up getting used. It seems this just sets everything to an empty string. Poked around with the `flag` package and can't figure out what I am missing here.

jaredpar: Oh so this is parsing the command line params later on?

dagood: I think looking at the other way to use `flag` first makes it a little clearer--you can pass `flag.StringVar` a pointer to insert the parsed arg value into:

```
var foo string
flag.StringVar(&foo, ...)
flag.Parse()
if foo == "hi" {
...
```

`flag.String` just allocates the memory internally rather than having you pass in your own pointer. You end up with `*string` rather than `string` so `flag.Parse()` can change the value.

qmuntal (approved):

eng/pipelines/release-build-pipeline.yml
```diff
@@ -0,0 +1,249 @@
...
+      microsoftGoInternalPipeline: 958
+      # Set a variable for each polling parameter. If a variable starts off with 'nil', it will be
+      # overridden with a logging command as the pipeline moves on.
+      MicrosoftGoPRNumber: ${{ parameters.poll1MicrosoftGoPRNumber }}
+      MicrosoftGoCommitHash: ${{ parameters.poll2MicrosoftGoCommitHash }}
+      MicrosoftGoBuildID: ${{ parameters.poll3MicrosoftGoBuildID }}
+      MicrosoftGoImagesPRNumber: ${{ parameters.poll4MicrosoftGoImagesPRNumber }}
+    steps:
+      - script: |
+          # Catch and prevent whole-job retries. Retrying the whole job after a
```
chsienki, line 78: This is cool. Does AzDO have a notion of 'child' pipelines? Essentially a release would consist of 4 different sub-pipelines, which feed results into the next one?  

I think Jenkins used to have something like that, so you could then retry the individual steps. Not blocking just something to think about.

chsienki (approved):

eng/pipelines/release-build-pipeline.yml
dagood: Well, when you have multiple jobs or stages, AzDO does let you retry just the ones that failed. In theory, if all we needed were incremental retries, this could be a single pipeline with individual jobs (or stages) for each retryable chunk. But:

* There's no way to change the go-infra commit run by a retry. (E.g. if a quick pipeline fix is needed before a retry will work.)
* Can't change any variables/parameters. (E.g. if the dev submits their own PR and wants the pipeline to poll that instead of the old, dead PR number.)
  * AzDO UI does let you toggle stages to run/not run when you queue the build, so instead of retrying, you could queue a build, change the parameters, and toggle off the stages that shouldn't re-run, but this is cumbersome.
* Acquiring an agent can take a while. Splitting up jobs multiplies that time.

Some years ago I tried release pipelines and they seemed more flexible for these kind of mid-release changes. But there is no YAML workflow, and they are called "classic" at this point, so best not to add new dependencies on them.

I'll put this in the design doc (in <https://github.com/microsoft/go-infra/pull/43>)--will be useful to have this context around in case AzDO adds new features in the area. 😄

PR #43, opened 2022-05-20
Add doc for running a release with automation

dagood:
Add instructions for the pipeline being added:

* https://github.com/microsoft/go-infra/pull/42

Rendered: https://github.com/dagood/go-infra/blob/dev/dagood/release-build-usage/docs/release-process/instructions.md


docs/release-process/instructions.md
```diff
@@ -0,0 +1,85 @@
...
+
+Once the values are filled in, press Run.
+
+You can close the tab as soon as the build starts looking for an agent. If nothing goes *terribly* wrong, the build will notify you if it needs intervention by commenting on the release issue.
+
+## microsoft/go revision
+
+Sometimes, we have to release when upstream hasn't made any changes. For example, to fix an issue with the FIPS implementation, or Go being built incorrectly by our infrastructure.
+
+In this case, run release-build, and increment the revision number from what it was for the latest release. If 1.18.3-1 was the latest, use 1.18.3-2 as the version.
```
dagood, line 32: This actually needs more intervention with the current pipeline. It doesn't update MICROSOFT_REVISION, so the build asset JSON will be wrong, resulting in incorrect URLs and go-images tag names.

* https://github.com/microsoft/go/issues/575

I'm going to work on this now, and we can hold off on merging the doc until then.

Similar goes for any boring branch release (pre 1.19):

* https://github.com/microsoft/go/issues/470

docs/release-process/instructions.md
```diff
@@ -0,0 +1,85 @@
...
+
+If dealing with multiple releases at the same time (for example, 1.17 and 1.18 security patches released on the same day), create one issue per version number. This makes it easier to keep track of each release's progress as an outside observer.
+
+> The "release-start" pipeline is planned to create these issues automatically for common servicing events. (See [README.md](README.md).)
+
+## Upstream release
+
+After an upstream release, or in anticipation of an upstream release, go to the [`microsoft-go-infra-release-build`](https://dev.azure.com/dnceng/internal/_build/results?buildId=1780270&view=results) pipeline. Press "Run new" and fill in the blanks. For example, a 1.42.3-1 patch release would look like this:
+
+> ![](images/run-upstream-release.png)
```
jaredpar, line 15: Where do you end up explaining the purpose of the other parameters here?

dagood: Hmm, I was thinking the field titles would explain it fine, but the numbered ones are definitely not clear at this point in the doc. I'll add a description.

Do you think the version number, issue number, and variable group ones are clear enough from the screenshot?

qmuntal (approved):

docs/release-process/instructions.md
dagood: Added a list below the screenshot explaining a little more.

docs/release-process/instructions.md
```diff
@@ -0,0 +1,92 @@
...
+
+> The "release-start" pipeline is planned to create these issues automatically for common servicing events. (See [README.md](README.md).)
+
+## Upstream release
+
+After an upstream release, or in anticipation of an upstream release, go to the [`microsoft-go-infra-release-build`](https://dev.azure.com/dnceng/internal/_build/results?buildId=1780270&view=results) pipeline. Press "Run new" and fill in the blanks in the popup.
+
+For example, if upstream is releasing 1.42.3, the parameters might look like this:
+
+> ![](images/run-upstream-release.png)
```
chsienki, line 17: These don't show up on the github preview, is that just an artifact of the rendering?

chsienki (approved):

docs/release-process/instructions.md
dagood: Hmm, looks like they don't show up in the "rich diff" mode, but they do show up at "view file". Strange, don't think I've noticed that before. Seems like GitHub could have at least shown some indication that there's an image if there's some good reason they can't simply show it inline. 😕

PR #45, opened 2022-06-03
Support `fips-linux` variant and `main` for microsoft/go-images

dagood:
This lets us build and auto-update Docker images for the `microsoft/main` branch and give them reasonable tag names. Also add support for a `fips-linux/*` variant that wraps the `*` tag with an extra layer that enables the OpenSSL FIPS experiment by default.

The PR also lets us do this if we choose (included in the example build):
* https://github.com/microsoft/go/issues/418

Example build: https://dev.azure.com/dnceng/internal/_build/results?buildId=1803638&view=results

A few examples:
```
golangpublicimages.azurecr.io/dev/oss/go/microsoft/golang:main-bullseye-amd64
golangpublicimages.azurecr.io/dev/oss/go/microsoft/golang:main-fips-bullseye-amd64
golangpublicimages.azurecr.io/dev/oss/go/microsoft/golang:main-cbl-mariner1.0-amd64
golangpublicimages.azurecr.io/dev/oss/go/microsoft/golang:main-fips-cbl-mariner1.0-amd64
```

### DIFF
--- a/buildmodel/buildmodel.go
+++ b/buildmodel/buildmodel.go
@@ -329,7 +329,7 @@ func joinTag(s ...string) string {
 	if len(s) == 0 {
 		return ""
 	}
-	b := strings.Builder{}
+	var b strings.Builder
 	first := true
 	for i := 0; i < len(s); i++ {
 		if s[i] == "" {
--- a/buildmodel/buildmodel.go
+++ b/buildmodel/buildmodel.go
@@ -17,6 +17,7 @@ import (
 	"github.com/microsoft/go-infra/buildmodel/dockermanifest"
 	"github.com/microsoft/go-infra/buildmodel/dockerversions"
 	"github.com/microsoft/go-infra/goversion"
+	"github.com/microsoft/go-infra/stringutil"
 )
 
 // ReadJSONFile reads one JSON value from the specified file.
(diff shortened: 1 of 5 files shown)


buildmodel/buildmodel.go
```diff
@@ -275,3 +321,25 @@ func makeOSArchPlatform(os, osVersion string, env *dockerversions.ArchEnv) *dock
...
 }
+
+// joinTag joins the given strings with "-" to form a Docker tag (or partial tag). Empty strings are
+// ignored and do not result in extra "-" characters. This is especially useful when it would be
+// inconvenient for the caller to keep track of which elements might be an empty string.
+func joinTag(s ...string) string {
+	if len(s) == 0 {
+		return ""
+	}
+	b := strings.Builder{}
```
qmuntal, line 332: nit: normal variable declaration is preferred over shorthand syntax when declaring a variable with default initialization.

```suggestion
	var b strings.Builder
```

qmuntal (approved):

buildmodel/buildmodel.go
dagood: Good catch. I think what's happening is that I type `x := `, look in the package for a `NewX` func if I haven't used `X` in a while, end up seeing that `X` is expected to be created directly, and keep moving forward with `{}` rather than revising the start of the line. 😄 Hopefully now I can keep an eye out for it better.

qmuntal: This exact same usability issue happens to my every now and then, but once my eyes see this pattern, they command the fingers without going through the brain 😄

chsienki (approved):

PR #51, opened 2022-07-05
Attach retry instructions to release automation build as a markdown doc

dagood:
* Resolves https://github.com/microsoft/go/issues/606. Added some notes to the issue about other approaches. I think for now, this is a good enough improvement.

Upload a doc describing how to retry to the AzDO build as a "summary" of the build. This makes AzDO display the subset-of-markdown document its own tab on the build summary screen, which is a lot easier to get to than digging into the logs of a specific build step to find the instructions. This is also an opportunity to filter out unnecessary instructions.

Update release docs to match.

Include a direct link to the page showing the retry doc in status update release issue comments.

Fix an inconsistent "." in the Report Failure step vs. the Report Success step.

---

The doc looks like this in the build:

image

And the comment on the release issue looks like this:

> [1857900](https://dev.azure.com/dnceng/internal/_build/results?buildId=1857900): :rotating_light: Failed to run microsoft/go build prep steps for 1.18.3-1.
> [Click here to see 1857900 retry instructions.](https://dev.azure.com/dnceng/internal/_build/results?buildId=1857900&view=ms.vss-build-web.run-extensions-tab)

Changes to the repo-side instructions doc are here: https://github.com/dagood/go-infra/blob/dev/dagood/retry-doc/docs/release-process/instructions.md#retrying


azdo/azdo.go
```diff
@@ -79,6 +79,13 @@ func SetPipelineVariable(name, value string) {
 	fmt.Printf("##vso[task.setvariable variable=%v]%v\n", name, value)
 }
 
+// UploadBuildSummary uses an AzDO logging command to upload a build summary. When used on a
+// markdown file, the rendered document shows up on the build page in an "Extensions" tab. The path
+// must be a full path.
+func UploadBuildSummary(path string) {
```
jaredpar, line 85: This is more generic than build, can be used to upload anything. Consider naming `UploadSummary`

jaredpar (approved):

qmuntal (approved):

azdo/azdo.go
dagood: Hmm, what I originally had in mind for the name was that it's just specifying "AzDO build" for a little clarify at the point where the func is used. Similar to `GetBuildWebURL`. (But `upload a build summary` in the comment ruins that.)

For `task.setvariable` I had the same thing in mind, but I called it `SetPipelineVariable(name, value string)`. Not consistent vs. `Build`.

Maybe it'd get the point across better to make it explicit and stick more closely to the `task.setvariable`/`task.uploadsummary` names: `LogCmdSetVariable`, `LogCmdUploadSummary`. It would also make it easier to recognize that this isn't a network call. (The lack of `ctx context.Context` and return value seems odd if you think it's a network call.) I'm doing this rename.

PR #60, opened 2022-08-23
fuzzcrypto initial implementation

qmuntal:
This PR implements the command `fuzzcrypto`, which runs a curated list of fuzz tests sequentially, i.e. from a CI pipeline.

This functionality might be built in Go fuzzing framework in a future release, but as of today it does not allow fuzzing multiple targets: https://github.com/golang/go/issues/46312.

`fuzzcrypto` also supports running fuzz tests from other projects, e.g. [go-cose](https://github.com/veraison/go-cose), giving us real-life coverage in addition to our synthetic fuzz tests. 

The initial list of fuzz test covers the following packages: `crypto/hash`, `crypto/hmac`, `crypto/rsa`, and `crypto/ecdsa` (via go-cose).


cmd/fuzzcrypto/std/rsa_test.go
```diff
@@ -0,0 +1,164 @@
...
+		hashed := sha256.Sum256(msg)
+		sig, err := rsa.SignPSS(rand.Reader, key, crypto.SHA256, hashed[:], &opts)
+		if err != nil {
+			t.Fatal(err)
+		}
+		err = rsa.VerifyPSS(&key.PublicKey, crypto.SHA256, hashed[:], sig, &opts)
+		if err != nil {
+			t.Error(err)
+		}
+		hashed[0] ^= 0xff
```
dagood, line 132: In theory we could improve the fuzzing domain by having the fuzzer pass us non-zero `[sha256.Size]byte` arrays to xor the entire checksum against, right? (Well, don't necessarily have to use xor for that, but I think you get what I mean. 😄)

I don't know how this extra domain would be *useful* to fuzz, though, personally. (What would it be, some kind of hash collision in the underlying library? And the extra time it would take to run each iteration might mean we don't catch some other issue in the parts of the domain that matter....)

dagood (approved):
A suggestion, couple questions, but LGTM for an initial implementation.

cmd/fuzzcrypto/main.go
```diff
@@ -0,0 +1,105 @@
...
+		"-fuzz", "^"+fuzzname+"$", // ensure we are strictly matching name
+	)
+	cmd.Dir = filepath.Join(".", dir)
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	err := cmd.Run()
+	if err != nil {
+		if _, ok := err.(*exec.ExitError); !ok {
+			return err
+		}
```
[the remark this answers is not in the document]
qmuntal: Good point, I'll change the error handling logic to exit with status code 1 when there is at least on failing test. The Go fuzz framework writes failed cases in `$testpackage/testdata/$testname`, so it should be possible to know what failed by either scanning stdout or by running `git status`.

cmd/fuzzcrypto/main.go
```diff
@@ -0,0 +1,105 @@
...
+		sumweights += t.weight
+	}
+
+	for i, t := range targets {
+		d := time.Duration((t.weight / sumweights) * float64(*fuzztime))
+		log.Printf("Running fuzz target %s for %v. %d/%d completed\n", t.name, d, i, len(targets))
+
+		err := fuzz(t.name, d)
+		if err != nil {
+			log.Printf("Railed to run fuzz target %s: %v\n", t.name, err)
```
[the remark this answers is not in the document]
qmuntal: Will do that. I've also added the flag `-v` for verbose logging, false by default. Fuzz test std outputs will only be piped to the cli output if `-v` is set.

cmd/fuzzcrypto/std/rsa_test.go
qmuntal: RSA sign tests already have too many parameters, it takes almost 1 minute to start converging! so I'm reluctant to another slice input. Additionally, all the verification logic is outside the Go stack, so the fuzzing engine won't learn anything by xoring bits, the xor array would be completely random.

On the other hand, your point is valid, the hash collision coverage is low with the current setup. I'll implement a compromise: add a new `uint8` input which will be used to decide which byte of the hash to xor and with which value. At least we will not be xoring always the first byte with `0xff`.

cmd/fuzzcrypto/main.go
```diff
@@ -32,6 +32,7 @@ nor the fuzz seeding time.
 `
 
 func main() {
+	var verbose = flag.Bool("v", false, "verbose logging")
 	var fuzztime = flag.Duration("fuzztime", 5*time.Minute, "total fuzzing time")
 	var run = flagRegex("run", "regex matching a set of fuzz targets")
```
dagood, line 37: I forgot to ask, the plan was to have it so that the fuzz tests are built in public CI but only run in internal CI. I assume this would need a new build-only flag here?

(For the first iteration I think it's fine to leave out the public CI side, because it's not super important that it's stable until the work is finished and we stop looking at the new pipeline as often.)

dagood (approved):

cmd/fuzzcrypto/main.go
qmuntal: The plan is to run `go build ./...` in public CI to spot compilation errors before running tests. There wouldn't be any actual test execution there, so there is no need to have a build-only flag.

cmd/fuzzcrypto/main.go
```diff
@@ -0,0 +1,113 @@
...
+	return cmd.Run()
+}
+
+// filterTargets returns the list of fuzz targets to execute,
+// which is either equal or a subset of alltargets.
+func filterTargets(run *regexp.Regexp) []target {
+	if run == nil {
+		return alltargets
+	}
+	targets := make([]target, 0, len(alltargets))
```
chsienki, line 106: Nit: Is there a reason we have to have `alltargets` in its own file? It took me a minute to understand where it was even getting them from. Consider putting them in the main file upfront.

chsienki (approved):

cmd/fuzzcrypto/main.go
dagood: `go build ./...` is good for the tool, but it ignores `*_test` files, right?

But I suppose `go test ./...` in `cmd/fuzzcrypto` would try to build `std` and catch issues (and still not run fuzz tests because this mode only cares about "normal" tests). There's still the question of `go-cose` though... since it's in its own Go module dir, seems like it'd need another build/test command.

cmd/fuzzcrypto/main.go
dagood: My two cents: finding symbols in different files in the package is something that I had trouble with when coming to Go from C# (where "one file per class" is pretty common style that worked well for me). I'm still calibrating my sense of when to split Go files up. 😄 This one seemed reasonable to me because:

* The list of targets is essentially data that the main command uses. Implementation vs. configuration.
* Acyclic dependency between the files. Cycles are fine in this context, but it seems like a good thing to try for because it's easier to navigate (at least, not as bad as cycles), and the file can be torn off into its own package if need be.
* The target list will probably get a decent number of changes that don't affect the main command at all.

(I'm also curious what @qmuntal thinks about this!)

cmd/fuzzcrypto/main.go
qmuntal: > go build ./... is good for the tool, but it ignores *_test files, right?

Good catch. Check this new commit: 817fb7c598d765ba5c6201bde19ae3649327077d. I think it will solve the we-need-to-run-fuzz-as-if-they-were-tests problem. You can now pass the number of iterations to `fuzztime` (it still supports a duration) by suffixing the value with `x`, so running `go run . -fuzztime 1x` will effectively run each target one time, even `go-cose` targets. Additionally, there is no way we found a security bug with `1x`, as we will only be running the first item of the seed corpus.

cmd/fuzzcrypto/main.go
qmuntal: I put `alltargets` in another file because, as @dagood has mentioned, it more a configuration thing than implementation logic, which could eventually be passed as a json file in the cli parameters.

I used to love the "on file per class" approach when C# was my language of choice, but after being abducted by Go I tend to fit as many symbols as the code reviewer would allow me in the same file. All have its pos and cons, won't enter in detail here 😄

cmd/fuzzcrypto/main.go
```diff
@@ -0,0 +1,153 @@
...
+
+The total fuzzing time, set by -fuzztime, is distributed among the executed fuzz tests.
+fuzzcrypto might run longer than -fuzztime as it does not take into account build time
+nor the fuzz seeding time.
+`
+
+func main() {
+	var verbose = flag.Bool("v", false, "verbose logging")
+	var fuzzDuration durationOrCountFlag
+	flag.Var(&fuzzDuration, "fuzztime", "total time to spend fuzzing or number of iterations that the fuzz target will be executed before exiting; default is to run for 5min")
```
dagood, line 39: I think the `-help` doc should mention the syntax, maybe just copy upstream's (with our own default) now that we use the same syntax?

```
-fuzztime t
    Run enough iterations of the fuzz target during fuzzing to take t,
    specified as a time.Duration (for example, -fuzztime 1h30s).
	The default is to run forever.
    The special syntax Nx means to run the fuzz target N times
    (for example, -fuzztime 1000x).
```

dagood (approved):

cmd/fuzzcrypto/main.go
qmuntal: Done!

PR #61, opened 2022-08-25
Add YAML quirk and style notes

dagood:
I found myself copying the same note to two places affected by the same quirk, and decided to write a little more about it here instead. Also added a few more notes. It's less cramped than writing in a YAML comment. 😄

There are a few weasel words where I'm not confident I'm right and I don't have the right tools to truly understand how AzDO is interpreting things.

https://github.com/microsoft/go-infra/blob/dev/dagood/pipeline-quirks/docs/pipeline-yml-style.md


qmuntal (approved):

docs/pipeline-yml-style.md
```diff
@@ -0,0 +1,97 @@
...
+
+Useful AzDO Pipeline YAML docs:
+* [YAML schema reference for Azure Pipelines](https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema)
+  * [Shortcut steps: `pwsh`, `script`, `publish`, etc.](https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/steps?view=azure-pipelines)
+* [Template types & usage](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/templates?view=azure-devops)
+* [Build and release tasks](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/?view=azure-devops)
+
+## One stage per job
+
+If a job fails in a running AzDO build with only one stage, you must wait for every single job to fail before you can retry the failed job. This is a problem: if a long job fails early but other long jobs are still running, we waste time waiting for the whole stage to finish.
```
chsienki, line 15: I'm not sure I understand this, it seems to contradict itself. You say on stage per job is bad, then say the workaround is to have one stage per job?

chsienki (approved):
LGTM minus the job section

docs/pipeline-yml-style.md
dagood: Reading this again, nasty ambiguity here:

> If a job fails in a running AzDO build with only one stage ...

With what I had in mind at the time, it means this:

> If a job [out of many other jobs] fails in a running AzDO build [and the build has] only one stage [that contains all the jobs] ...

But could also mean:

> If a job fails in a running AzDO build [and that job has] one stage ...

I think I'll replace this with an example to cut out the English. 😄

chsienki: Ahh ok, I understand now. Possibly that was made worse by me not understanding exactly what a job vs a stage is. The replacement explanation is even better though :)

docs/pipeline-yml-style.md
```diff
@@ -0,0 +1,112 @@
...
+- ${{ each value in parameters.thing }}:
+  - ${{ insert }}: ${{ value }}
+    answer: 42
+```
+
+We edit the YAML files with generic YAML tools.
+
+## Indentation style
+
+The YAML files in this repository indent when starting a list, which isn't typical for pipeline YAML. This isn't specific to AzDO, but
```
dagood, line 84: Just noticed this--it kinda works as a lead-in for the example, but that wasn't what I intended to have here. 😄 

```suggestion
We indent when starting a new list in the YAML files in this repository. This isn't typical for pipeline YAML: in AzDO docs, a list's `-` is on the same column as the element containing the list. Both styles are commonly found in general YAML files. For example, we use:
```

PR #63, opened 2022-09-14
Add Fuzz test pipeline and validation in PR CI

dagood:
Adds an internal fuzz testing pipeline. Adds a job in in go-infra PR validation to run fuzz testing with `1x` to validate that each go-infra PR doesn't break the tool or tests, but without actually running with any new fuzz data.

Internal build with a simulated test failure to see that the testdata is included in the uploaded artifact: https://dev.azure.com/dnceng/internal/_build/results?buildId=1994508&view=results

I also have a full run going:
https://dev.azure.com/dnceng/internal/_build/results?buildId=1994517&view=results

The artifact also contains the source code and go-cose submodule data, but I think this is fine: we don't expect this pipeline to fail often, and it turns out to be 2 MB. To improve it, we would have to search the dir and copy the testdata files to a fresh dir to upload, and that doesn't seem worth the complexity.

The fuzz test currently grabs the newest Go version built by microsoft/main. We could pretty reasonably extend this to release branches, or even consume our publicly available bits rather than pulling from the official build pipeline. This would require listing each release branch here and customizing it further for pre-1.19 (no GOEXPERIMENT, different toolset) so I'm not doing this in the initial implementation at least.

This PR doesn't yet support notifications, filing issues, etc., it just makes the pipeline fail.

### DIFF
--- a/eng/pipelines/README.md
+++ b/eng/pipelines/README.md
@@ -5,23 +5,25 @@ This directory contains Azure DevOps (AzDO) YAML pipelines for CI and utilities.
 Public pipeline definitions using these YAML files:
 
 * [`pr-pipeline.yml`](pr-pipeline.yml) - PR validation
-  * [`microsoft-go-infra`](https://dev.azure.com/dnceng/public/_build?definitionId=1051)
+  * 🚀 dnceng-public [`microsoft-go-infra`](https://dev.azure.com/dnceng-public/public/_build?definitionId=197)
 
 Internal pipeline definitions:
 
 * [`update-images-pipeline.yml`](update-images-pipeline.yml) - Update dependencies in microsoft/go-images after a Go build.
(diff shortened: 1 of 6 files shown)


qmuntal (approved):
LGTM and (L)

chsienki (approved):

qmuntal:
100minutes might not be enough, it just gives around 5min per target. We should either increase `fuzztime` or run different targets in parallel (different jobs?) using `-run` to group targets.

dagood:
100 minutes is actually the infra timeout, fuzztime is only 60m right now. 🙁 I think it would be reasonable to increase to something like 3h for now.

> run different targets in parallel (different jobs?) using `-run` to group targets.

Hmm, I think we could split up the targets among N jobs similar to how we split the fuzztime. (So we can just kick off N jobs, rather than directly specifying which job runs which fuzz tests.) I want to give that a try in a followup PR.

PR #64, opened 2022-09-23
Add fuzz target bucketization, more jobs

dagood:
We aren't sure we have enough fuzz time with one agent running through each target sequentially over an hour, so this PR distributes the targets between several jobs. (https://github.com/microsoft/go-infra/pull/63#issuecomment-1248463682)

Distributing the targets between agents is generally more effective than running every target on multiple agents because Go fuzz testing builds up a local cache of interesting data points that it uses to refine its search for issues, guided by coverage.

This PR doesn't fix the existing issues we've been having with memory usage. These seem to actually be issues in the code under test, not the infra, so we can/will fix these as followup and hopefully end up with green fuzz test runs.

Test build I triggered: https://dev.azure.com/dnceng/internal/_build/results?buildId=2002188&view=results

### DIFF
--- a/cmd/fuzzcrypto/main.go
+++ b/cmd/fuzzcrypto/main.go
@@ -4,6 +4,7 @@
 package main
 
 import (
+	"container/heap"
 	"flag"
 	"fmt"
 	"log"
@@ -12,6 +13,7 @@ import (
 	"path"
 	"path/filepath"
 	"regexp"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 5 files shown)


cmd/fuzzcrypto/main.go
```diff
@@ -95,6 +117,32 @@ func flagRegex(name, usage string) **regexp.Regexp {
...
+		if err != nil {
+			return fmt.Errorf("string before '/' is not a number: %v", err)
+		}
+		c, err = strconv.Atoi(after)
+		if err != nil {
+			return fmt.Errorf("string after '/' is not a number: %v", err)
+		}
+
+		if b < 1 || b > c {
+			return fmt.Errorf("bucket %v is not in range 1 - %v", b, c)
```
qmuntal, line 139: Mathematical notation might be clearer here:

```suggestion
			return fmt.Errorf("bucket %v is not in range [1, %v]", b, c)
```

eng/pipelines/jobs/fuzz.yml
```diff
@@ -18,46 +20,56 @@ parameters:
...
+            parameters:
+              platform: ${{ parameters.platform }}
+
           - script: |
-              cd cmd\fuzzcrypto
-              set GOEXPERIMENT=cngcrypto
-              go run . -v -fuzztime ${{ parameters.fuzztime }}
+              $(cdFuzzCrypto)
+              go env
+              $(setExperiment)${{ parameters.goExperiment }}
```
qmuntal, line 59: Could we use AzDO `env` instad of setting the experiment via bash or cmd notation?
Lik this:

```yaml
- script: |
    go env
    go run ...
  displayName: Fuzz
  env: ${{ parameters.goExperiment }}

```

qmuntal (changes requested):

eng/pipelines/jobs/fuzz.yml
dagood: True. Also `workingDirectory` to get rid of the other variable.

cmd/fuzzcrypto/main.go
dagood: Good idea. Can also call it an interval to point out the notation usage. 😄

eng/pipelines/jobs/fuzz.yml
dagood: Ah right, `env` entries don't show up in the build logs. 😬 I usually stick with scripts with `set -x` for reproducibility/comprehensibility. (Although only on Linux.)

In this case `env` is perfectly fine because `go env` prints out the value anyway.

qmuntal (approved):

chsienki (approved):

PR #68, opened 2022-10-15
List all builds in single release tracking issue

dagood:
* Resolves https://github.com/microsoft/go/issues/605
* Now just one release tracking issue per release day instead of 3.
* Add a queue-time parameter to the "start" pipeline that lets the release dev plug in their GitHub username to have the bot subscribe them automatically to the release day issue.
* Update the instructions.

I think a reasonable way to review this is to start by looking at the golden files to understand what I'm aiming for. Then, review `buildreport`, which has all the text generation and GitHub sync logic. Then, the rest is integration with `releasego` and the YAML.

Here's an example of what it would look if we were only releasing 1.19 and the auto-update PR hit patch conflicts and can't merge:

> image

* Example with more builds completed: https://github.com/dagood/go-release-test/issues/66

---

Making sure we don't lose data if there's a concurrent update attempt was a challenge, and I ended up using `git push` to the project wiki as the locking mechanism. See comments in `buildreport.UpdateIssueBody` for more details about why the GitHub API didn't work for this.

GitHub does allow you to restrict wiki edits to accounts with write access to the repo. Otherwise, someone could add data to the wiki that that the bot would then copy to the tracking issue. I wanted to make `report` confirm that the wiki is in restricted mode before fetching data from it, but unfortunately that option isn't visible in the GitHub API.

I think this (or something like it) is the best we can do until we go ahead and start creating our own infra on Azure, where auth is my concern.

I've written a default wiki page for microsoft/go that mentions we aren't using it for more information and gives some pointers to more info: https://github.com/microsoft/go/wiki.


buildreport/buildreport.go
```diff
@@ -0,0 +1,454 @@
...
+				"confirm the MAR/MCR images are updated using commands like these:\n\n" +
+				"```\n" +
+				"image=mcr.microsoft.com/oss/go/microsoft/golang:1-bullseye\n" +
+				"docker pull $image\n" +
+				"docker run -it --rm $image go version\n" +
+				"```\n"
+		}
+	}
+
+	if notification != "" {
```
qmuntal, line 216: nit: if you negate this condition then you can return early and reduce one indentation level.

buildreport/buildreport.go
```diff
@@ -0,0 +1,454 @@
...
+	if source.Name != "" {
+		s.Name = source.Name
+	}
+	if source.URL != "" {
+		s.URL = source.URL
+	}
+	if source.Status != "" {
+		s.Status = source.Status
+	}
+	if source.LastUpdate != (time.Time{}) {
```
qmuntal, line 278: The idiomatic way to do this check is using [time.IsZero()](https://pkg.go.dev/time#Time.IsZero).

buildreport/buildreport.go
```diff
@@ -0,0 +1,454 @@
...
+		s.LastUpdate = source.LastUpdate
+	}
+	if source.StartTime != (time.Time{}) {
```
qmuntal, line 281: The idiomatic way to do this check is using [time.IsZero()](https://pkg.go.dev/time#Time.IsZero).

buildreport/buildreport.go
```diff
@@ -0,0 +1,454 @@
...
+					b.WriteString(" ([Retry](")
+					b.WriteString(r.URL)
+					b.WriteString("&view=ms.vss-build-web.run-extensions-tab))")
+				}
+			}
+		}
+		b.WriteString(" | ")
+		b.WriteString(r.Status)
+		b.WriteString(" | ")
+		if r.StartTime != (time.Time{}) {
```
qmuntal, line 411: The idiomatic way to do this check is using [time.IsZero()](https://pkg.go.dev/time#Time.IsZero).

buildreport/buildreport.go
```diff
@@ -0,0 +1,454 @@
...
+			b.WriteString(r.StartTime.Format("2006-01-02 15:04 MST"))
+		}
+		b.WriteString(" | ")
+		if r.LastUpdate != (time.Time{}) {
```
qmuntal, line 415: The idiomatic way to do this check is using [time.IsZero()](https://pkg.go.dev/time#Time.IsZero).

buildreport/buildreport_test.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+}
+
+func Test_commentBody_body(t *testing.T) {
+	exampleTime, err := time.Parse(time.RFC3339, "2012-03-28T01:02:03Z")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	r := func(version, id, pipeline string, status string) State {
```
qmuntal, line 56: brrrr.... `s/r/newState`

buildreport/buildreport_test.go
```diff
@@ -0,0 +1,141 @@
...
+		b := State{
+			Version:    version,
+			ID:         id,
+			Name:       pipeline,
+			URL:        "https://example.org/",
+			Status:     status,
+			StartTime:  exampleTime,
+			LastUpdate: exampleTime.Add(time.Minute * 5),
+		}
+		return b
```
qmuntal, line 66:
```suggestion
		return State{
			Version:    version,
			ID:         id,
			Name:       pipeline,
			URL:        "https://example.org/",
			Status:     status,
			StartTime:  exampleTime,
			LastUpdate: exampleTime.Add(time.Minute * 5),
		}
```

qmuntal (changes requested):

buildreport/buildreport.go
dagood: Good spot, it also lets me return the error value from `githubutil.Retry` directly.

buildreport/buildreport.go
dagood: Thanks, definitely looks nicer.

I mentioned in standup that doing this change made it behave differently, but it's just because I thoughtlessly slapped in `if source.LastUpdate.IsZero()` rather than `if !source.LastUpdate.IsZero()`. 😄 Fixed.

`IsZero` does seem to be a little different because it compares actual time rather than directly comparing the `location` pointers, but this is a good thing for our sanity, and it doesn't change any behavior I care about.

qmuntal (approved):

buildreport/buildreport.go
```diff
@@ -0,0 +1,453 @@
...
+// Update updates the report then sends a notification comment if necessary.
+func Update(ctx context.Context, owner, repoName, pat string, issue int, s State) error {
+	if err := UpdateIssueBody(ctx, owner, repoName, pat, issue, s); err != nil {
+		return err
+	}
+	return Notify(ctx, owner, repoName, pat, issue, s)
+}
+
+// UpdateIssueBody updates the given issue with new state. Requires the target GitHub repo to have
+// the wiki activated to perform safer concurrent updates than a simple issue description edit.
```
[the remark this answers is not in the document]
dagood: I did a simple retry at first, but no. 😄 I spawned some goroutines to run edits at the same time and it hit issues immediately.

The reason GitHub is able to detect an edit conflict in the web UI is because the issue description editor is a form, and it uses some kind of form validation token/data that somehow stores info about the "current" state of the issue as far as that client is aware. If the form's "current" data doesn't match the authoritative current data on GitHub, GitHub refuses to run the update.

The GitHub API doesn't send any "current" state, only the target body text. So GitHub has no way to tell if one API-based edit conflicts with another, and the last one wins.

Passing an ETag (["If-Match"](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Match)) or date ("If-Unmodified-Since") header are the normal ways to make this work in REST APIs, but neither works here. GitHub's ETags are weak and can't be used this way, and "If-Unmodified-Since" does nothing at all.

I wasn't able to find *any* GitHub API that has concurrency safety that we could reasonably use. The GitHub API that *should* behave like a safe Git push isn't even safe and silently performs the equivalent of a force push. I assume this is because it would be expensive/hard to make any concurrent-safe API work with a distributed system, so they didn't. 😅 

But once we go outside the GitHub API, `git push` works fine. I didn't want to spam everyone with new `ref`s on the main microsoft/go repo, and the wiki happens to also be a Git repo with permissions that are linked to microsoft/go permissions, so it seemed reasonable.

jaredpar (approved):

PR #70, opened 2022-10-20
Add config file with minimum tool version

dagood:
* Resolves https://github.com/microsoft/go/issues/538
* Resolves https://github.com/microsoft/go/issues/749
  * (Added this in here rather than creating a new PR because it's relatively simple but would conflict with the changes here.)

This config file would let us specify a minimum version of the tool to make sure everyone's up to date, but also let other teams use the tool because now they could specify their own submodule, patches, and status file dirs.


dagood:
A recent Git update in Ubuntu broke our tests: https://bugs.launchpad.net/ubuntu/+source/git/+bug/1993586

patch/patch.go
```diff
@@ -87,3 +91,81 @@ func WalkPatches(dir string, fn func(string) error) error {
...
+		Config:  conventionalConfig,
+		RootDir: dir,
+	}, nil
+}
+
+func isDir(dir string) (ok bool, err error) {
+	if info, err := os.Stat(dir); err != nil {
+		return info != nil && info.IsDir(), err
+	}
+	return true, nil
```
jaredpar, line 152: Won't passing a normal file path, not a directory, end up hitting this line?

dagood: Yeah, looks like I put the `info != nil && info.IsDir()` in the wrong place. Will fix it up. (I think naming the variable `dir` even though we don't know it's a dir yet makes it confusing, so I'll change that too.)

qmuntal (approved):
nit: there are multiple errors and log traces using the `%v` verb to format file paths. It will be confusing when the path contains spaces, would be good to quote them using `%q` or `%#q` verbs.

dagood:
The general way I try to approach that, in case we have different styles: in the middle of a line, I use `%#q` to make sure spaces are unambiguous, but at the end of a line--especially after a `:`--just `%v` seems clear. (I vaguely remember a common tool doing it this way, but don't remember what it was.) For paths, I prefer `%#q` over `%q` so Windows paths end up with `\` rather than `\\`.

I do see a few places I didn't follow that properly, will fix.

jaredpar (approved):

cmd/git-go-patch/rebase.go
```diff
@@ -50,5 +53,39 @@ func handleRebase(p subcmd.ParseFunc) error {
...
+	}
+
+	if err := warnIfOutsideSubmodule(goDir); err != nil {
+		// Just log for diagnosis, this warning isn't critical.
+		fmt.Printf("Unexpected error while checking wd: %v\n", err)
+	}
+	return nil
+}
+
+func warnIfOutsideSubmodule(submoduleDir string) error {
```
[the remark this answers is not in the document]
dagood: I (slightly) prefer it this way because it lets me write ordinary error handling:

```go
	if err != nil {
		return fmt.Errorf("unable to get working dir: %v", wd)
	}

	if err != nil {
		fmt.Printf("Unexpected error while checking wd: unable to get working dir: %v\n", wd)
		return
	}
```

Generally speaking, it's idiomatic to have an `error` return most of the time, so to me, it's easier to read if the code leans into that.

I'll switch it over to see the difference for real. @qmuntal, also curious what your thoughts on this are.

PR #72, opened 2022-10-24
Add "ExtractAsAuthor" git-go-patch config option

dagood:
* Resolves https://github.com/microsoft/go/issues/383

Also reduces log length a little by only logging one time before moving patches from the temp dir to the real dir. Improve error handling if something happens during this process so we don't lose diagnosability.

Tested round trip in microsoft/go, and it seems to work fine. Adding the config file and configuring it to set author to the bot also worked as expected.

### DIFF
--- a/patch/patch.go
+++ b/patch/patch.go
@@ -5,14 +5,17 @@
 package patch
 
 import (
+	"bufio"
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"log"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"sort"
+	"strings"
 
 	"github.com/microsoft/go-infra/executil"
 )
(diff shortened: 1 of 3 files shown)


chsienki (approved):

patch/patch.go
```diff
@@ -92,6 +95,76 @@ func WalkPatches(dir string, fn func(string) error) error {
...
+		} else if strings.HasPrefix(line, subjectPrefix) && subject.Len() == 0 {
+			subject.WriteString(line[len(subjectPrefix):])
+			subject.WriteString("\n")
+		} else {
+			subject.WriteString(line)
+			subject.WriteString("\n")
+		}
+	}
+	h.Subject = subject.String()
+	h.Content = content.String()
```
dagood, line 147: If it's ever a problem to store the patch file in memory (so far, < 1MB), a few options:

* I was poking around some other potential patch tool improvements and found `git mailinfo`: https://git-scm.com/docs/git-mailinfo. It splits a patch file into multiple files plus stdout that we could re-assemble into the final file.
* We can do something similar ourselves, in-process.

jaredpar (approved):

PR #74, opened 2022-10-27
Avoid spurious updates in "git-go-patch extract"

dagood:
* Resolves https://github.com/microsoft/go/issues/751

Rendered test data docs: https://github.com/microsoft/go-infra/tree/dev/dagood/skip-effective-unchanged/patch/testdata

---

For example, take this patch conflict resolution commit: https://github.com/microsoft/go/commit/566a95f0ab053e4cf1648b20b3b95f5fcf581158. When I used `git go-patch extract`, it changed patch file 0001, too, not just the 0003 and 0005 that needed fixes. I had to manually revert the spurious 0001 change. The 0001 diff looked like this:

```diff
diff --git a/patches/0001-cmd-dist-add-JSON-output-support-for-some-tests.patch b/patches/0001-cmd-dist-add-JSON-output-support-for-some-tests.patch
index f007b02af4..20548be539 100644
--- a/patches/0001-cmd-dist-add-JSON-output-support-for-some-tests.patch
+++ b/patches/0001-cmd-dist-add-JSON-output-support-for-some-tests.patch
@@ -17,10 +17,10 @@ in JSON format, and the ordinary logs are still important.
  1 file changed, 10 insertions(+), 1 deletion(-)

 diff --git a/src/cmd/dist/test.go b/src/cmd/dist/test.go
-index 759377ffa54568..595e618cf29763 100644
+index 1663ff8c94868e..1ce8546bde668d 100644
 --- a/src/cmd/dist/test.go
 +++ b/src/cmd/dist/test.go
-@@ -31,6 +31,7 @@ func cmdtest() {
+@@ -29,6 +29,7 @@ func cmdtest() {

        var noRebuild bool
        flag.BoolVar(&t.listMode, "list", false, "list available tests")
@@ -28,7 +28,7 @@ index 759377ffa54568..595e618cf29763 100644
        flag.BoolVar(&t.rebuild, "rebuild", false, "rebuild everything first")
        flag.BoolVar(&noRebuild, "no-rebuild", false, "overrides -rebuild (historical dreg)")
        flag.BoolVar(&t.keepGoing, "k", false, "keep going even when error occurred")
-@@ -57,6 +58,7 @@ type tester struct {
+@@ -55,6 +56,7 @@ type tester struct {
        msan        bool
        asan        bool
        listMode    bool
...
```

The changes are to the index hash (related to the last commit that updated the file that was patched) and the line numbers of the chunks that are modified. None of these changes are necessary to successfully apply the patch file, and none of them have any effect on the result: a `git diff` between the result of applying the "old" patch and "new" patch returns zero diff.

This PR essentially adds logic to run the `git diff` on the results and keep the old patch if there is no diff. This feature means that after I fix up patch conflicts and run `git go-patch extract`, I get https://github.com/microsoft/go/commit/566a95f0ab053e4cf1648b20b3b95f5fcf581158 without doing any more manual work.

This also means that if there aren't any patch conflicts, `git go-patch apply` -> `git go-patch extract` will round-trip properly without modifying any patch files. This also means you can `git go-patch rebase`, edit an individual commit, and only that specific patch file will be changed after `git go-patch extract`.

This feature is all-or-nothing per patch file. Chunks within each modified patch file might still have spurious changes. Editing out unnecessary changes to chunk line numbers seems like it could work (`am`/`apply` is smart and will find the correct place to make the change), but would require us to parse patch files *much* deeper.

---

Spurious changes don't always show up. The reason for the spurious changes is updating the submodule pointer to a later commit without regenerating the patch files. (We routinely do this in microsoft/go--the ability to do this is what makes the project reasonably maintainable.)

`git am` is smart enough to be able to apply old patch files to new submodule commits, but `git format-patch` has no knowledge of the existing patch files and the old commit they might be based on, so it chooses fresh index and chunk line numbers based on the current submodule commit.

---

Unfortunately, this process is slow, because it involves a lot of Git commands run in sequence. On my machines, the "reducing spurious changes" stopwatch reads 2.172s on Linux and 9.879s on Windows. Before this feature, extracting patches took 50-200ms. I think the time is worth it. Other teams might have (many) more patches than we do, so I included a built-in timer and a `-verbatim` flag to let people disable the feature if necessary. The `-verbatim` flag is also useful to bypass the feature in case it makes something go terribly wrong.

I haven't truly investigated speeding it up. I'm skeptical that more can be done in parallel, because we're limited by `am` and running patches in sequence on a single cloned Git repo. We could potentially find a Git library that has less overhead than running many `git` processes. We could also use the existing submodule instead of making a fresh clone, which increases the risk of destroying work-in-progress but reduces copying. `am` may also have more options to make it faster directly.


patch/matching.go
```diff
@@ -0,0 +1,131 @@
...
+	if m.gitDir, err = gitcmd.NewTempCloneRepo(submodulePath); err != nil {
+		return nil, fmt.Errorf("failed to create temp clone of submodule to evaluate patch changes: %v", err)
+	}
+	if err := gitcmd.Run(m.gitDir, "checkout", "-q", baseCommit); err != nil {
+		return nil, fmt.Errorf("failed to check out base commit in temp repo: %v", err)
+	}
+	// Associate each patch's header (identity, as far as this comparison is concerned) with the
+	// patch file's path. This lets us try old and new when we find a new patch's header, later.
+	m.existingPatchPathsByHeader = make(map[Header]string)
+	if err := WalkPatches(patchesDir, func(path string) error {
```
jaredpar, line 43: Nit: may want to update docs for `WalkPatches` to clarify it's returned in a deterministic fashion. The internal docs indicate this but may help in the summary. Always suspicious of anything that involves the file system.

patch/matching.go
```diff
@@ -0,0 +1,131 @@
...
+		p, err := ReadFile(path)
+		if err != nil {
+			return err
+		}
+		absPath, err := filepath.Abs(path)
+		if err != nil {
+			return err
+		}
+		m.existingPatchPathsByHeader[p.Header] = absPath
```
jaredpar, line 52: Did you consider an assert / check that this key is not present yet? It shouldn't be ...

cmd/git-go-patch/extract.go
```diff
@@ -177,18 +192,39 @@ func handleExtract(p subcmd.ParseFunc) error {
...
+		writeNewPatch := true
+
+		// Now that we're done modifying p, see if it has any effective differences vs. the old
+		// patch with the same header.
+		if matcher != nil {
+			matchingStopwatch.Start()
+			if err := matcher.Apply(path, p); err != nil {
+				return failSuggestVerbatim("failed to check patch for changes", err)
+			}
+			if matcher.LastApplyExistingMatch != "" {
```
jaredpar, line 204: Rather than `LastApplyExistingMatch` being temporal state on the type, why not return the old path from `Apply`? It only seems to be used in this one place

dagood: This is kind of based on `bufio.Scanner`--call one method to advance, and then others to check what it found afterwards. But that isn't necessarily a good pattern, and since there's only one state field and it's always used, returning it does sound better. Will change (and give it a name to fit).

patch/matching.go
dagood: I didn't, yeah, we should fail in that case. Someone could *potentially* try to split up a patch without changing the subject/author/date, just a different diff and filename, and I believe the current code would end up overwriting every one of the split files with the content of the last file we hit in the walk.

jaredpar (approved):

chsienki (approved):

dagood:
Thanks. 🙂

PR #84, opened 2023-05-16
Handle upstream 1.21.0 change to have zero in tag

dagood:
* Fixes https://github.com/microsoft/go/issues/888
* Handle https://github.com/golang/go/issues/57631

Upstream has this set up to apply at 1.21.0, so put that in our infra. This is called by the part of release automation that polls the upstream repo on release day. If it doesn't end up working out for some reason, we'll need to revert this PR and re-run automation.

I considered polling both the old and new formats at the same time, but I would prefer to simply get it right rather than add a tag check race. I haven't seen any indication that upstream would go back to the old versioning scheme if something went wrong, so I don't know if it would actually end up being used even in that case.

### DIFF
--- a/goversion/goversion.go
+++ b/goversion/goversion.go
@@ -109,15 +109,26 @@ func (v *GoVersion) Full() string {
 	return v.MajorMinorPatchPrereleaseRevision() + v.NoteWithPrefix()
 }
 
-// UpstreamFormatGitTag returns the version in the format upstream uses for Git tags. Specifically,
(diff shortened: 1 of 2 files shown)


qmuntal (approved):

dagood:
> If it doesn't end up working out for some reason, we'll need to revert this PR and re-run automation.

I realized this isn't true: if this change goes wrong, we can simply create a release update PR manually and feed that into automation for the next attempt. (And then fix it later for the next release.) Very low risk.

PR #86, opened 2023-06-12
Fix FuzzRSASignPSS

qmuntal:
`FuzzRSASignPSS` is failing on the fuzzing pipeline because the *allowed* error messages have changed in go1.20. What used to be reported as `crypto/rsa: key size too small for PSS signature` is now the typed error `rsa.ErrMessageTooLong` (see [CL 443195](https://go-review.googlesource.com/c/go/+/443195)).

### DIFF
--- a/cmd/fuzzcrypto/std/rsa_test.go
+++ b/cmd/fuzzcrypto/std/rsa_test.go
@@ -123,9 +123,12 @@ func FuzzRSASignPSS(f *testing.F) {
 		hashed := sha256.Sum256(msg)
 		sig, err := rsa.SignPSS(rand.Reader, key, crypto.SHA256, hashed[:], &opts)
 		if err != nil {
-			switch err.Error() {
 			// Key size being too small is an invalid input, not a problem.
+			switch err.Error() {
 			case "crypto/rsa: key size too small for PSS signature", "crypto/rsa: invalid key size":
+				return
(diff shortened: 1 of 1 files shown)


cmd/fuzzcrypto/std/rsa_test.go
```diff
@@ -123,11 +123,14 @@ func FuzzRSASignPSS(f *testing.F) {
 		hashed := sha256.Sum256(msg)
 		sig, err := rsa.SignPSS(rand.Reader, key, crypto.SHA256, hashed[:], &opts)
 		if err != nil {
-			switch err.Error() {
 			// Key size being too small is an invalid input, not a problem.
+			switch err.Error() {
 			case "crypto/rsa: key size too small for PSS signature", "crypto/rsa: invalid key size":
```
dagood, line 128: I take it the idea here is to leave `"crypto/rsa: key size too small for PSS signature"` in place for back compat even though it's a typed error now?

dagood (approved):

cmd/fuzzcrypto/std/rsa_test.go
qmuntal: Yep, just in case we want to fuzz an older Go version.

jaredpar (approved):

PR #97, opened 2024-03-01
Use aka.ms Rest API instead of Arcade

qmuntal:
We are currently installing dotnet and Arcade in the release pipeline to create the aka.ms links.

We could achieve the same by directly using the aka.ms Rest API, which is documented here: https://microsoft.sharepoint.com/teams/RedirectionServiceInfo/SitePages/Redirection%20Service%20API%20Guide.aspx#_Toc532312225.

The benefits are clear: we can remove all the Arcade idiosyncrasy from the repo and the code becomes easier to understand/maintain.


dagood:
I kicked off https://dev.azure.com/dnceng/internal/_build/results?buildId=2391791&view=results to test it, iterated a few times. (It's safe to run it again multiple times if you use "run new" on that page to keep the params. It's set up with a different aka.ms URL prefix for testing.)

I fixed a few things, but now it's hitting a 400 error, so I'm stopping for the time being.

qmuntal:
It's working now 😄

qmuntal:
> > It's working now 😄
> 
> Well.. not yet.

Ok, now it works again. I had to implement the `akams.CreateOrUpdateBulk` so the pipeline doesn't fail when rerunning with the the same parameters, in which case we want to update the link, not create a new one.

Still WIP, I want to add some more tests for the `akams` package.

dagood (approved):

eng/pipelines/release-build-pipeline.yml
```diff
@@ -251,17 +250,15 @@ jobs:
...
-
                 releasego akams \
                   -build-asset-json '$(buildAssetJsonFile)' \
                   -prefix '$(VanityUrlPrefix)' \
-                  /p:AkaMSClientId='$(akams-client-id)' \
-                  /p:AkaMSClientSecret='$(akams-client-secret)' \
-                  /p:AkaMSTenant='$(AkaMSTenant)' \
-                  /p:AkaMSCreatedBy='$(AkaMSCreatedBy)' \
-                  /p:AkaMSGroupOwner='$(AkaMSGroupOwner)'
+                  -clientID='$(akams-client-id)' \
```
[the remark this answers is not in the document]
qmuntal: Personal preference, but I don't have any strong reason for using `-foo=bar` over `-foo bar`. In fact I now realize that the later is the norm in this repo, so I'll use it here too.

internal/akams/akams.go
```diff
@@ -140,49 +172,101 @@ func (c *Client) CreateOrUpdateBulk(ctx context.Context, links []CreateLinkReque
...
+	resp, err := c.httpClient.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("request failed: %v", err)
+	}
+	return resp, nil
+}
+
+// chunkSlice chunks s into bulkSize chunks preserving the order of the items,
+// encodes each chunk into JSON, and calls fn with the encoded chunk.
+// The maximum encoded size of each chunk is limited to maxSizeBytes.
```
dagood, line 224: I started putting together a suggestion to clarify that the first line doesn't just mean "into N chunks" where N = bulkSize, but ended up moving some other things around to (IMO) make it more digestible:

```suggestion
// chunkSlice chunks s, encodes each chunk into JSON,
// and calls fn with the encoded chunk.
// The maximum encoded size of each chunk is limited to maxSizeBytes,
// and each chunk has at most bulkSize items.
// The order of the items is preserved.
```

internal/akams/akams.go
```diff
@@ -140,49 +172,101 @@ func (c *Client) CreateOrUpdateBulk(ctx context.Context, links []CreateLinkReque
...
+		if err := enc.Encode(s[i]); err != nil {
+			return err
+		}
+		buf.Truncate(buf.Len() - 1) // Remove the trailing newline added by enc.Encode.
+		buf.WriteByte(',')          // Add a comma to separate items.
+		if buf.Len() > maxSizeBytes {
+			// The last item was too big.
+			if size := buf.Len() - lastSize; size > maxSizeBytes {
+				// The last item is too big to fit in a chunk.
+				return fmt.Errorf("item %d is too large: %d bytes", i, size)
```
dagood, line 250: Include max size to help the user see if they've accidentally configured this to something super small?
```suggestion
				return fmt.Errorf("item %d is too large: %d bytes > %d byte maximum", i, size, maxSizeBytes)
```

dagood (approved):
🚀

gdams (approved):

internal/akams/akams_test.go
```diff
@@ -0,0 +1,325 @@
...
+
+func TestCreateBulk_Fail_MaxSize(t *testing.T) {
+	links := []akams.CreateLinkRequest{
+		{ShortURL: "short", TargetURL: "target", CreatedBy: "cb"},
+	}
+	close, client := setup(t)
+	defer close()
+	client.SetBulkLimit(0, 1)
+	err := client.CreateBulk(context.Background(), links)
+	wantErr := "item 0 is too large: 91 bytes"
```
dagood, line 118:
```suggestion
	wantErr := "item 0 is too large: 91 bytes > 1 byte maximum"
```

internal/akams/akams_test.go
```diff
@@ -0,0 +1,325 @@
...
+
+func TestUpdateBulk_Fail_MaxSize(t *testing.T) {
+	links := []akams.UpdateLinkRequest{
+		{ShortURL: "short", TargetURL: "target"},
+	}
+	close, client := setup(t)
+	defer close()
+	client.SetBulkLimit(0, 1)
+	err := client.UpdateBulk(context.Background(), links)
+	wantErr := "item 0 is too large: 74 bytes"
```
dagood, line 163:
```suggestion
	wantErr := "item 0 is too large: 74 bytes > 1 byte maximum"
```

dagood (approved):
Whoops, I broke the tests. 🙂

PR #100, opened 2024-03-15
Implement 1ES PT for production pipelines

dagood:
Add 1ES PT to the production pipelines. These were pretty straightforward.

Test runs:
* release-go-start: https://dev.azure.com/dnceng/internal/_build/results?buildId=2405183&view=results
  * Filed https://github.com/dagood/go-release-test/issues/67 (private repo) and updated it correctly.
  * This kicked off release-build: https://dev.azure.com/dnceng/internal/_build/results?buildId=2405185&view=results, which (correctly) waited around because the version I gave it (1.24.0) doesn't exist.
* release-build https://dev.azure.com/dnceng/internal/_build/results?buildId=2405187&view=results
  * Failed to publish new aka.ms links for 1.21.8-1 with a dev prefix. This is almost definitely due to secret rotation, not this PR. https://github.com/microsoft/go-lab/issues/61
* release-go-images: https://dev.azure.com/dnceng/internal/_build/results?buildId=2405186&view=results
  * Correctly waited around for my private testing repo to get a PR to update to 1.24.0-1, which never happened.

These tests aren't all that extensive, but given what I expect about the kind of operations that 1ES PT affects, I'd be satisfied merging this and working on the aka.ms link issue afterwards.

The aka.ms link generation test does cover some good ground. It demonstrates successfully downloading the build asset JSON from an official build, which was something I was concerned might be limited in some way or need more work to migrate.

I noticed that https://github.com/microsoft/go/pull/1156 is missing the BuildAssets build asset JSON pipeline artifact, so I didn't run tests against that yet. I probably missed it during the switch, because artifact publishing had to change for 1ES PT.

Reviewing with "ignore whitespace" highly recommended.


eng/pipelines/release-build-pipeline.yml
```diff
@@ -65,230 +65,249 @@ variables:
...
+      sourceAnalysisPool:
+        name: NetCore1ESPool-Svc-Internal
+        image: 1es-windows-2022
+        os: windows
+      tsa:
+        enabled: true
+        configFile: $(Build.SourcesDirectory)/eng/compliance/tsaoptions.json
+
+    stages:
+      - template: stages/releasego.yml
```
qmuntal, line 88: For the sake of minimizing changes in this PR, what about keeping `releasego.yml` unmodified and define the stage in here instead?

Something like this:

```yml
extends:
  template: v1/1ES.Official.PipelineTemplate.yml@1ESPipelineTemplates
  parameters:
    ...
    stages:
    - stage: Release
       jobs:
       - template: jobs/releasego.yml
```

qmuntal (changes requested):

gdams (approved):
Changes LGTM but we should address Quim's comment before merging

eng/pipelines/release-build-pipeline.yml
dagood: > - template: jobs/releasego.yml

In case it comes up later: this is the case where you need to use `- template: /eng/pipelines/jobs/releasego.yml@self`, because a relative string here is evaluated relative to the `v1/1ES.Official.PipelineTemplate.yml@1ESPipelineTemplates` that's being extended.

dagood: Good call, the repetition in each pipeline by declaring the stage is quite minimal, and the diff is much clearer.

dagood:
New test builds:
* https://dev.azure.com/dnceng/internal/_build/results?buildId=2405745&view=results
* https://dev.azure.com/dnceng/internal/_build/results?buildId=2405736&view=results
* https://dev.azure.com/dnceng/internal/_build/results?buildId=2405746&view=results

qmuntal (approved):

dagood:
> * Failed to publish new aka.ms links

This worked in my new test build! I see that .NET Engineering fixed it this morning.

PR #101, opened 2024-03-20
Add innerloop pipeline to the release process

dagood:
* For https://github.com/microsoft/go/issues/1159

Here's a basic approach to running innerloop tests in our release process. It puts the burden of the "join" on the release runner. (That "Approve" button now has a functional purpose. 🙂)

---

The build job *could* wait for one pipeline and then wait for the other in sequence, but this compromises on a few things:

* It would take longer to report the status of the second pipeline to the release runner if it fails.
* If the first pipeline hangs, it prevents reporting on the other pipeline.
* Retry queueing UX gets bad when both builds are in flight.
  * Copying *two* values is much worse than one. The Run dialog is modal and doesn't save progress, so the user would need to have a temp doc to transfer things over, or have multiple tabs open, or something along those lines.
  * Encoding two values into one copy-pasteable value is complicated to implement and makes it more difficult to specify manually when the dev is doing a fixup.
    * However, longer term, perhaps going all-in and using a YML/JSON object rather than individual text fields would be a good idea. (Has its own downsides.)

---

The build job could also be even more serialized: run innerloop to completion and only then queue the official job. (Or the other way around.) This avoids the compromises, but it's ~40min--not a pleasant thing to add synchronously, especially when there's the opportunity to run it parallel to signing, which can vary dramatically.

### DIFF
--- a/docs/release-process/instructions.md
+++ b/docs/release-process/instructions.md
@@ -38,11 +38,13 @@ For example, a security patch release for 1.18 and 1.19 may look like this:
     * It will take some time to reserve a build agent. Expect up to ten minutes.
 
 1. Wait for notifications on the release tracking issue.
-    * When the "microsoft-go-infra-release-build" build is complete and successful for every version in the release, continue to the next step.
+    * Continue once both of these builds complete successfully for every version in the release:
+        * **microsoft-go-infra-release-build**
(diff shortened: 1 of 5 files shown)


qmuntal (approved):
Nice!

gdams (approved):

PR #110, opened 2024-03-25
enable codeql

gdams:
Additional compliance requirement


dagood:
We should do AzDO-side CodeQL as a central pipeline, like https://github.com/microsoft/go/pull/1184.

PR #114, opened 2024-04-05
Implement ORAS helper command

qmuntal:
`azoras` is a tool that helps work with the [ORAS CLI](https://github.com/oras-project/oras) and Azure ACR.
The subcommands implement common workflows for image annotations and maintenance.

It currently implement three basic workflows:
- `azoras install`: Install the ORAS CLI from GitHub releases
- `azoras login`: Logs in the ORAS CLI to Azure ACR using the Azure CLI
- `azoras deprecate`: Adds a end-of-life annotation to the given image or list of images


cmd/azoras/azoras.go
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+package main
```
dagood, line 3:
```suggestion

package main
```

cmd/azoras/deprecate.go
```diff
@@ -0,0 +1,133 @@
...
+	go run ./cmd/azoras deprecate images.txt -bulk -eol 2022-12-31T23:59:59Z
+		`,
+		Handle:         handleDeprecate,
+		TakeArgsReason: "The fully qualified image to deprecate or a file containing a newline-separated list of images to deprecate if -bulk is set.",
+	})
+}
+
+func handleDeprecate(p subcmd.ParseFunc) error {
+	eolStr := flag.String("eol", "", "The end-of-life date for the image in RFC3339 format. Defaults to the current time.")
+	bulk := flag.Bool("bulk", false, "Deprecate multiple images.")
```
dagood, line 42: It seems like it would be simpler to accept the file the value of this flag--then the non-flag args only have one meaning.

In general the func could be arranged as accumulating a list of images to mark (regardless of what input they're from) and then processing a list, to keep the flow a bit simpler.

Related: from a usage perspective, it seems odd not to accept multiple args. I think not emitting an error if there are more/less args is the most concrete issue--but simplifying seems like it might naturally make that issue go away.

cmd/azoras/deprecate.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+	for _, prev := range prevs {
+		t, err := time.Parse(time.RFC3339, prev)
+		if err == nil && t.Before(eol) {
+			// The image is already deprecated.
+			log.Printf("Image %s is already past its EoL date of %s\n", image, prev)
+			return nil
+		}
+	}
+	cmdOras := exec.Command("oras", "attach", "--artifact-type", artifactTypeLifecycle, "--annotation", annotationNameEoL+"="+eol.Format(time.RFC3339), image)
```
dagood, line 101: Hefty line, I'd probably do this:

```suggestion
	cmdOras := exec.Command(
		"oras", "attach",
		"--artifact-type", artifactTypeLifecycle,
		"--annotation", annotationNameEoL+"="+eol.Format(time.RFC3339),
		image)
```

internal/archive/tar.go
```diff
@@ -0,0 +1,35 @@
...
+	tr := tar.NewReader(r)
+	for {
+		header, err := tr.Next()
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return nil, err
+		}
+		if header.Name == name {
```
dagood, line 30: I believe this handles multiple entries with the same name differently than the `tar` tool's defaults (keeping last) so could end up with different results. 😄 I think detecting a duplicate and returning an error would be reasonable because I can't imagine the tool being packed this way intentionally by the authors, and it's something a human would almost definitely miss when updating the ORAS version otherwise.

cmd/azoras/README.md
```diff
@@ -0,0 +1,44 @@
...
+Optionally, azoras can install the ORAS CLI for you running:
+
+```shell
+azoras install
+```
+
+## Authentication
+
+azoras doesn't handle authentication, you need to login to your Azure ACR registry with the ORAS CLI.
+See https://oras.land/docs/how_to_guides/authentication for more information.
```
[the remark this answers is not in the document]
qmuntal: Good point, I'll rephrase this section.

cmd/azoras/install.go
```diff
@@ -0,0 +1,178 @@
...
+
+// checksums is copied from https://github.com/oras-project/oras/releases/download/v1.1.0/oras_1.1.0_checksums.txt.
+var checksums = map[string]string{
+	"oras_1.1.0_linux_s390x.tar.gz":  "067600d61d5d7c23f7bd184cff168ad558d48bed99f6735615bce0e1068b1d77",
+	"oras_1.1.0_windows_amd64.zip":   "2ac83631181d888445e50784a5f760f7f9d97fba3c089e79b68580c496fe68cf",
+	"oras_1.1.0_darwin_arm64.tar.gz": "d52d3140b0bb9f7d7e31dcbf2a513f971413769c11f7d7a5599e76cc98e45007",
+	"oras_1.1.0_linux_armv7.tar.gz":  "def86e7f787f8deee50bb57d1c155201099f36aa0c6700d3b525e69ddf8ae49b",
+	"oras_1.1.0_linux_amd64.tar.gz":  "e09e85323b24ccc8209a1506f142e3d481e6e809018537c6b3db979c891e6ad7",
+	"oras_1.1.0_linux_arm64.tar.gz":  "e450b081f67f6fda2f16b7046075c67c9a53f3fda92fd20ecc59873b10477ab4",
+	"oras_1.1.0_darwin_amd64.tar.gz": "f8ac5dea53dd9331cf080f1025f0612e7b07c5af864a4fd609f97d8946508e45",
```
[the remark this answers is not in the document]
qmuntal: Makes sense.

cmd/azoras/install.go
```diff
@@ -0,0 +1,178 @@
...
+	if path, err := exec.LookPath("oras"); err == nil {
+		log.Printf("ORAS CLI is already installed at %s\n", path)
+		return nil
+	}
+	dir, err := installDir()
+	if err != nil {
+		return err
+	}
+	if err := downloadAndInstallOras(dir); err != nil {
+		os.RemoveAll(dir) // Best effort to clean up the directory.
```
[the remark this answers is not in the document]
qmuntal: Yep, that would be unfortunate. Removing this code.

cmd/azoras/deprecate.go
qmuntal: I like your suggestion, will do!

dagood:
Moved [to our internal repo](https://dev.azure.com/dnceng/internal/_git/microsoft-go-infra?version=GBdev/qmuntal/azoras&path=/cmd/azoras) for now while we work on a pipeline etc.

PR #115, opened 2024-04-05
Fix failing fuzz tests

qmuntal:
This PR fixes the failing fuzz tests that I've identified as "legitim" failures. There might be more, this is a first step to reduce the noise in our continuos fuzzing pipeline (https://dev.azure.com/dnceng/internal/_build?definitionId=1182&_a=summary).

I've bumped the go-cose submodule to the latest commit, as it contains a fix for https://github.com/veraison/go-cose/issues/165, which was detected in our pipeline.

### DIFF
--- a/cmd/fuzzcrypto/std/rsa_test.go
+++ b/cmd/fuzzcrypto/std/rsa_test.go
@@ -11,6 +11,7 @@ import (
 	"crypto/sha256"
 	"errors"
 	"math/big"
+	"strings"
 	"testing"
 )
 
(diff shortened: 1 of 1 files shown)


cmd/fuzzcrypto/std/rsa_test.go
```diff
@@ -171,6 +170,26 @@ func FuzzRSASignPKCS1v15(f *testing.F) {
 	})
 }
 
+func validRSAError(err error) bool {
```
dagood, line 173: Maybe call this `expectedRSAError`? To me an error can be "valid" (e.g. self-consistent), and still show up in unexpected circumstances.

It might be better to be more specific though... even in this file, this func is being used for some `err`s and not others, so there are still *some* places where it's expected vs. unexpected. Maybe:
`if errorCorrectlyRejectedRSAInput(err) { return }`

dagood (approved):

PR #130, opened 2024-06-24
Dev/mertakman/add update azure linux command

mertakman:


cmd/releasego/update-azure-linux.go
```diff
@@ -0,0 +1,98 @@
...
+	"regexp"
+
+	"github.com/microsoft/go-infra/githubutil"
+	"github.com/microsoft/go-infra/subcmd"
+)
+
+func init() {
+	subcommands = append(subcommands, subcmd.Option{
+		Name:        "update-azure-linux",
+		Summary:     "Update the Azure Linux Microsoft Go version after release.",
```
gdams, line 19:
```suggestion
		Summary:     "Update the Go spec files for Azure Linux.",
```

gdams:
@mertakman you'll need to rebase this now :)

githubutil/githubutil.go
```diff
@@ -114,22 +114,41 @@ func FetchEachPage(f func(options github.ListOptions) (*github.Response, error))
...
+
 	return err
 }
 
-// FileExists is a function that will check if a file exists in a given repository.
-func FileExists(ctx context.Context, client *github.Client, owner, repo, filePath string) (bool, error) {
-	_, _, resp, err := client.Repositories.GetContents(ctx, owner, repo, filePath, nil)
-	if err != nil {
+// DownloadFile is a function that will download a file from a given repository.
+func DownloadFile(ctx context.Context, client *github.Client, owner, repo, branch, path string) (file []byte, exists bool, err error) {
```
dagood, line 131: I actually think this should return `file []byte, err error`, and return `nil, githubutil.ErrNotExists` if it doesn't exist. Then, the name `DownloadFile` matches the behavior better: callers that actually want to download the file only need to check `err != nil`. Only callers that care *why* the file couldn't be downloaded then need to check `errors.Is(err, githubutil.ErrNotExists)`. This also fits in with APIs like `os.ReadFile`.

mertakman: You're right, I'll make sure that I'm implementing this.

cmd/releasego/update-azure-linux.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+
+	return assets, nil
+}
+
+func downloadFileFromRepo(ctx context.Context, client *github.Client, owner, repo, branch, filePath string) ([]byte, error) {
+	fileContent, err := githubutil.DownloadFile(ctx, client, owner, repo, branch, filePath)
+	if err != nil {
+		if errors.Is(err, githubutil.ErrNotExists) {
+			// Handle the specific case of the file not existing
```
dagood, line 113: Why is this necessary? Could this be more generic and just wrap the error from `githubutil.DownloadFile` to convey this information?

cmd/releasego/update-azure-linux.go
```diff
@@ -0,0 +1,249 @@
...
+			return nil, fmt.Errorf("file '%s' not found in repository '%s' on branch '%s'", filePath, repo, branch)
```
dagood, line 114: I think this should use `%q` or `%#q` for escaping, to make it easier to spot weird strings.

cmd/releasego/update-azure-linux.go
```diff
@@ -0,0 +1,249 @@
...
+		} else {
+			// Handle other errors (network issues, authentication, etc.)
+			return nil, fmt.Errorf("failed to download file '%s': %w", filePath, err)
+		}
+	}
+
+	if len(fileContent) == 0 {
+		return nil, fmt.Errorf("downloaded file %s is empty", filePath)
```
dagood, line 122: It seems odd for this func to check this, because it's not necessarily a problem in general cases. I think it would make sense for `updateSignatureFile` etc. to check it.

cmd/releasego/update-azure-linux.go
```diff
@@ -0,0 +1,249 @@
...
+
+const (
+	golangSignaturesFilepath = "SPECS/golang/golang.signatures.json"
+	golangSpecFilepath       = "SPECS/golang/golang.spec"
+	cgManifestFilepath       = "cgmanifest.json"
+)
+
+var (
+	specFileGoFilenameRegex = regexp.MustCompile(`(%global ms_go_filename  )(.+)`)
+	specFileRevisionRegex   = regexp.MustCompile(`(%global ms_go_revision  )(.+)`)
```
dagood, line 135: I don't think we should rely on number of spaces. The number of spaces is just because the original author of the spec file aligned some virtual columns, so the number could increase or decrease if another human touches it at any point and wants to realign. I think it's perfectly fine if this code even ignores the number of spaces and always puts in a single space in the PR update result.

(In general, I think we should keep an eye out for changes to the spec file that would make automated tooling easier to write, less fragile, or easier to review. My understanding is that we're free to make changes like that to the spec file. (I did some of that in https://github.com/microsoft/azurelinux/commit/3d42d8a38cc9dfe64c8b695c287d06c7eedfca4d.))

cmd/releasego/update-azure-linux.go
```diff
@@ -0,0 +1,249 @@
...
+)
+
+func extractGoArchiveNameFromSpecFile(specContent []byte) (string, error) {
+	matches := specFileGoFilenameRegex.FindStringSubmatch(string(specContent))
```
dagood, line 139: I see other funcs have `specContent` as a `string` parameter, and I think that makes more sense than accepting `[]byte`.

cmd/releasego/update-azure-linux.go
```diff
@@ -0,0 +1,249 @@
...
+
+	if len(matches) < 2 {
+		return "", fmt.Errorf("no Go archive filename declaration found in spec content")
+	}
+
+	return strings.TrimSpace(matches[2]), nil
```
dagood, line 145: It's odd seeing `len(matches) < 2 { return }` then `matches[2]`... I think it works based on what I know about `FindStringSubmatch` (only returns `nil` or a fully filled out slice) but it catches my eye because it wouldn't be correct if the slice length actually varied.

I think checking for `nil` is normal with this API because it explicitly says it will return `nil`, not an empty slice or something like that. (I see this being done in golang/go, for one.)

cmd/releasego/update-azure-linux.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+
+	updatedContent := specFileRevisionRegex.ReplaceAllString(specContent, "${1}"+newRevisionVersion)
+	return updatedContent, nil
+}
+
+func updateSpecFile(buildAssets *buildassets.BuildAssets, signatureFileContent []byte) ([]byte, error) {
+	content := string(signatureFileContent)
+
+	_ = content
```
dagood, line 169: TODO?

cmd/releasego/testdata/update-azure-linux/golang_updated.spec
```diff
@@ -0,0 +1,163 @@
...
+%global gohostarch      amd64
+%endif
+%define debug_package %{nil}
+%define __strip /bin/true
+# rpmbuild magic to keep from having meta dependency on libc.so.6
+%define _use_internal_dependency_generator 0
+%define __find_requires %{nil}
+Summary:        Go
+Name:           golang
+Version:        1.22.4
```
dagood, line 17: This needs to be updated too.

cmd/releasego/update-azure-linux.go
```diff
@@ -0,0 +1,249 @@
...
+
+	return strings.TrimSpace(matches[2]), nil
+}
+
+func updateGoArchiveNameInSpecFile(specContent, newArchiveName string) (string, error) {
+	if !specFileGoFilenameRegex.MatchString(specContent) {
+		return "", fmt.Errorf("no Go archive filename declaration found in spec content")
+	}
+
+	updatedContent := specFileGoFilenameRegex.ReplaceAllString(specContent, "${1}"+newArchiveName)
```
dagood, line 153: In theory, I suppose we should escape `$` in newArchiveName, or maybe just error out. A `$` would never be in there intentionally though.

cmd/releasego/update-azure-linux_test.go
```diff
@@ -0,0 +1,102 @@
...
+	if err != nil {
+		t.Fatalf("Error reading spec file from path %s, error is:%s", signaturesFilePath, err)
+	}
+
+	updatedSignatureFile, err := updateSignatureFile(signaturesFile, "go1.22.4-20240604.2.src.tar.gz", path.Base(assets.GoSrcURL), assets.GoSrcHash)
+	if err != nil {
+		t.Errorf("Error updating CG Manifest file : %s", err)
+	}
+
+	updatedSignatureFile = append(updatedSignatureFile, '\n')
```
dagood, line 70: Why is this in test code rather than updateSignatureFile?

dagood (commented):
I think this has some WIP stuff in it, but I think most of the comments might still be useful. 😄 I do think `updateCGManifest` and `updateSignatureFile` themselves seem reasonable.

buildmodel/buildassets/buildassets.go
```diff
@@ -39,6 +39,8 @@ type BuildAssets struct {
 
 	// GoSrcURL is a URL pointing at a tar.gz archive of the pre-patched Go source code.
 	GoSrcURL string `json:"goSrcURL"`
+	// GoSrcHash is the SHA256 hash of the pre-patched Go source code tar.gz archive stored at GoSrcURL.
+	GoSrcHash string `json:"goSrcHash"`
```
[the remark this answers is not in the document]
mertakman: fixed

cmd/releasego/update-azure-linux.go
mertakman: fixed, Thanks!

cmd/releasego/update-azure-linux_test.go
```diff
@@ -0,0 +1,102 @@
...
+		t.Fatalf("Error updating Go archive name in spec file : %s", err)
+	}
+	updatedspecFile, err = updateGoRevisionInSpecFile(updatedspecFile, assets.GoVersion().Revision)
+	if err != nil {
+		t.Fatalf("Error updating Go revision in spec file : %s", err)
+	}
+
+	goldentest.Check(
+		t, "TestUpdateSpecFileContent ",
+		filepath.Join("testdata", "update-azure-linux", "golang_updated.spec"),
```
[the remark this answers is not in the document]
mertakman: fixed

cmd/releasego/update-azure-linux_test.go
mertakman: fixed!

gdams (approved):
LGTM, if we need to make further changes we can do so in a follow-up PR

PR #132, opened 2024-06-24
ci: run linter task on code changes

gdams:
I've made a few code changes to make the linter green but I'll defer to @dagood to confirm that these are sane


.github/workflows/linter.yml
```diff
@@ -0,0 +1,19 @@
+name: golangci-lint
+on:
+    push:
+      branches: [ "main" ]
+    pull_request:
+      branches: [ "main" ]
```
mertakman, line 6: Sometimes I'm opening a pull request on top of some other branch, I think we lose nothing if we handle this case too.

karianna: Agree that early warnings are good if it's a cheap cost to run them :-)

gitpr/gitpr.go
```diff
@@ -266,7 +266,7 @@ type GitHubRequestError struct {
 }
 
 func PostGitHub(ownerRepo string, request *GitHubRequest, pat string) (response *GitHubResponse, err error) {
-	prSubmitContent, err := json.MarshalIndent(request, "", "")
+	prSubmitContent, _ := json.MarshalIndent(request, "", "")
```
dagood, line 269: Not returning the err is simply a mistake. This is some early Go code before I was in tune with Go sensibilities. 😄 (I also wouldn't use bare `return` here--it's basically always more trouble than it's worth in my experience. No need to change that with this PR, though.)

dagood (commented):
We should include a config file in the repo so that it doesn't fall back to user settings, right?

gitpr/gitpr.go
gdams: updated PTAL

mertakman (approved):

.golangci.yml
```diff
@@ -0,0 +1,10 @@
+linters:
+  enable:
+    - errcheck
+    - gosimple
+    - govet
+    - ineffassign
+    - staticcheck
+    - unused
```
dagood, line 8: A few things that might be nice in this file:
* indicating which linters are default (probably unobjectionable/ordinary) and which ones we're specifically enabling
* a link to the ref docs.

.golangci.yml
```diff
@@ -0,0 +1,10 @@
+linters:
```
dagood, line 1: `golangci-lint run` fails on Windows with standard Git configuration due to CRLFs. See https://github.com/golangci/golangci-lint/issues/580. Forcing Git to use LF for Go files even on Windows is apparently the only fix, which is unfortunate, but it is something I've seen before in Go repos so I suppose people are used to it.

.golangci.yml
```diff
@@ -0,0 +1,10 @@
...
+  enable:
+    - errcheck
+    - gosimple
+    - govet
+    - ineffassign
+    - staticcheck
+    - unused
+    - gofumpt
+    - bodyclose
```
dagood, line 10: A few more that I think would be nice to add, but need some more changes so can be in a followup pr:

* `nakedret` (configured to not allow them at all, IMO)
* `predeclared`
* `unconvert` (may be controversial--perhaps a redundant convert is done for clarity)

I really want to use `goheader`, but it messes up docs: https://github.com/denis-tingaikin/go-header/issues/35.

dagood (approved):
I'll push changes for a couple of these comments because I was playing with it locally. LGTM. A few more rules caught my eye that could be added as followup.

PR #133, opened 2024-06-27
fix:gofumpt repo

mertakman:
Used https://github.com/mvdan/gofumpt for applying its better gofmt rules on repo.
At some point should consider if we should include gofumpt on our toolchain, it has really good rules that making Go code more readable.


buildreport/buildreport.go
```diff
@@ -104,7 +114,7 @@ func UpdateIssueBody(ctx context.Context, owner, repoName, pat string, issue int
 
 	startTime := time.Now()
 	for {
-		elapsed := time.Now().Sub(startTime)
+		elapsed := time.Since(startTime)
```
dagood, line 117: This is good, but it isn't from gofumpt, is it?

cmd/git-go-patch/extract.go
```diff
@@ -329,7 +331,7 @@ func (s *stopwatch) Start() {
 }
 
 func (s *stopwatch) Stop() {
-	s.elapsed += time.Now().Sub(s.start)
+	s.elapsed += time.Since(s.start)
```
dagood, line 334: Also not from gofumpt?

sync/sync.go
```diff
@@ -687,10 +687,10 @@ func MakeBranchPRs(f *Flags, dir string, entry *ConfigEntry) ([]SyncResult, erro
 		if force {
 			c.Args = append(c.Args, "--force")
 		}
+
 		c.Args = append(c.Args, auther.InsertAuth(remote))
-		for _, r := range refspecs {
-			c.Args = append(c.Args, r)
-		}
+		c.Args = append(c.Args, refspecs...)
+
```
dagood, line 693: Also doesn't seem like gofumpt.

dagood (approved):
Nice, I like these rules. There are a few changes that are good, but don't quite seem to technically belong in the gofumpt commit. (I don't have a problem with merging them anyway, because it's unlikely that will *actually* cause confusion in the future. With a squash merge, can just include in the description that a few cleanups were also done.)

sync/sync.go
mertakman: few linter errors are there I fixed with this PR.

PR #135, opened 2024-07-03
publish-announcement: sort versions to be decreasing order

gdams:


goversion/goversion.go
```diff
@@ -156,3 +156,40 @@ func extractPrerelease(part, prerelease *string) {
 		*part = (*part)[:i]
 	}
 }
+
+// IsNewerThan compares two versions and returns true if the receiver is newer than the argument
+// version. The comparison is done by comparing the major, minor, patch, revision, prerelease and note parts in that
+// order. If all parts are equal, false is returned. It may return an error if a part is not an integer.
```
dagood, line 162: I tried rewriting this to explain the intent behind "may return an error if a part is not an integer", but I'm ending up not really happy about exposing this method with it implemented this way. I see a few ways forward:

* Preprocessing the versions to actually make sure they're all integers. I think this makes more sense to the user, is more usable, and ends up with a cleaner implementation: https://gist.github.com/dagood/9d36726d640d3138165916636d564ecf
* We could move this back to publish-announcement and unexport it to remove the burden of making it a public API.
* This method could be improved to handle strings as well as numbers to remove `error` and simplify usage and the explanation. Basically: a more generic natural sort, rather than a natural sort *if* the data happens to meet some constraints.

mertakman: looks good!

dagood: @mertakman and I chatted about this and I liked the suggestion to use https://pkg.go.dev/sort#Interface and to simply put non-int-parseable elements at the end of the slice. This is easier to document as part of our exported API and is analogous to how Go sorts (or rather, doesn't sort?) various NaNs.

This is similar to the third alternative above, but we don't really need to make an effort to support "bad" versions in much detail. Also, using `sort.Interface` to make it more intuitive to use rather than the current func and `sort.Slice`. Perf isn't a significant factor here given our use cases.

(In the future we might want to consider restricting what `goversion` allows, to catch "bad" versions early and eliminate this edge case. From the dev/user's point of view, this would mean catching it in the microsoft-go-infra-release-start pipeline before the version number is used to do anything.)

dagood:
@mertakman I see a force push, but no changes to the sorting code in this PR--just a rebase or something so far?

goversion/goversion.go
```diff
@@ -156,3 +156,44 @@ func extractPrerelease(part, prerelease *string) {
...
+
+type GoVersions []*GoVersion
+
+func (versions GoVersions) Len() int      { return len(versions) }
+func (versions GoVersions) Swap(i, j int) { versions[i], versions[j] = versions[j], versions[i] }
+func (versions GoVersions) Less(i, j int) bool {
+	less := func(a, b string) bool {
+		intA, err := strconv.Atoi(a)
+		if err != nil {
+			return false
```
dagood, line 168: This doesn't result in "put non-int-parseable elements at the end of the slice" behavior--this test case fails, for example:

```
		{
			name: "sorting with non-int element",
			versions: GoVersions{
				New("1.2.5"),
				New("1.hi.4"),
				New("1.2.3"),
			},
			expected: GoVersions{
				New("1.2.5"),
				New("1.2.3"),
				New("1.hi.4"),
			},
		},
```

(The result is that the slice order doesn't change at all:
`expected 1.2.3 at index 1, got 1.hi.4`
`expected 1.hi.4 at index 2, got 1.2.3`)

It also isn't transitive:

`1.2.3` .Less `1.hi.2` is false
`1.hi.2` .Less `1.2.1` is false
So you'd expect `1.2.3` .Less `1.2.1` to be false, but it's true.
(I wrote a test case to find these: https://gist.github.com/dagood/9d55f83beb172425fdb45e05949492a2.)

I think this behavior is *ok* with a doc describing it. Leaving it unexported/internal seems even better to me though. I'll push a change to move it.

goversion/goversion.go
```diff
@@ -156,3 +156,44 @@ func extractPrerelease(part, prerelease *string) {
 		*part = (*part)[:i]
 	}
 }
+
+type GoVersions []*GoVersion
```
dagood, line 160: My concern with this and the old Less method is mainly that we're providing "weird" behavior in an exported API. But thinking about it again, we can just make it clear what "breaks" it.

```suggestion
// GoVersions implements [sort.Interface] and sorts versions in descending order.
// If Major, Minor, Patch, or Revision of any GoVersion in the slice can't be parsed by
// [strconv.Atoi], the result of using this type is undefined.
type GoVersions []*GoVersion
```

goversion/goversion_test.go
```diff
@@ -127,3 +128,76 @@ func TestGoVersion_UpstreamFormatGitTag(t *testing.T) {
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// Sort the versions
+			sort.Sort(tt.versions)
+			for i, v := range tt.versions {
+				if *v != *tt.expected[i] {
+					t.Errorf(tt.name)
```
dagood, line 197: As far as I know the subtest name is always in the log just above a test failure's output, so this isn't necessary.

```suggestion
```

cmd/releasego/testdata/publish-announcement/three-branches.golden.md
```diff
@@ -1,7 +1,7 @@
 ---
-post_title: Go 1.23.1-1, 1.22.8-1, and 1.21.11-16 Microsoft builds now available
+post_title: Go 1.22.8-1, 1.23.1-1, and 1.21.11-16 Microsoft builds now available
```
dagood, line 2: It looks like the logic in `generateBlogPostTitle` actually mangles the sort a little.

dagood (approved):
I pushed a few changes, now it LGTM. @mertakman or @gdams, take a look and make sure you agree and feel free to merge. 🙂

PR #141, opened 2024-07-10
Pass publishReleaseStudio to microsoft-go as optional parameter

dagood:
Using Release Studio means that publishing every single microsoft-go build to the public would be more wasteful than it was before. So, I'm planning on adding a `publishReleaseStudio` parameter to the microsoft-go build to enable publishing, and now it will only be enabled by the release automation flow (or by devs for testing).

This is tricky for compatibility. This PR adds a feature so that we can pass a parameter *if the variable is declared* by the target pipeline/branch. This means our release automation pipeline doesn't need to worry about whether or not each release branch or main branch has implemented `publishReleaseStudio` yet. It will simply pass true if it's declared.

This feature could be avoided by implementing `publishReleaseStudio` in lockstep across all branches. However, I think allowing for incremental progress is worthwhile, and this feature may be useful for other parts of release automation in the future, too.

---

You'd think there would be a simpler way to do this. A few ways I tried:

* Use a variable! This would work, but AzDO variables are very limited in functionality compared to parameters, and introduce other complexities/annoyances/etc.
* Have `publishReleaseStudio` default to `true` and have the CI triggers turn it off! I didn't see a way to determine whether a build is triggered by CI in a way that works while the template is being evaluated, like parameters do. Maybe there are some side effects that could be used to figure it out.

In general, I think complexity in Go code is cozy compared to complexity in AzDO yml code, so I don't think this is a difficult choice, but I thought it's worth mentioning. 😄

### DIFF
--- a/cmd/releasego/build-pipeline.go
+++ b/cmd/releasego/build-pipeline.go
@@ -14,6 +14,7 @@ import (
 	"log"
 	"net/http"
 	"strconv"
+	"time"
 
 	"github.com/microsoft/azure-devops-go-api/azuredevops/build"
 	"github.com/microsoft/go-infra/azdo"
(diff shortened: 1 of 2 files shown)


mertakman (approved):

cmd/releasego/build-pipeline.go
```diff
@@ -88,71 +110,161 @@ func handleBuildPipeline(p subcmd.ParseFunc) error {
 
 	// Make our own client. The AzDO library doesn't support the 7.1 API needed to pass parameters:
 	// https://docs.microsoft.com/en-us/rest/api/azure/devops/build/builds/queue?view=azure-devops-rest-7.1
-	client := new(http.Client)
+	var client http.Client
```
mertakman, line 113: try not to use default client, they are lacking of timeouts.

dagood: Good call. We haven't seen a timeout happen here, but there's a first time for everything, and it would hold up the release if it does.

PR #144, opened 2024-07-16
Generate aka.ms links for individually specified URLs

dagood:
Depends on https://github.com/microsoft/go-infra/pull/143

`releasego akams` needs to handle the individual URLs added to the assets.json file. This is implemented here with a fallback to the old way of determining these URLs.

`releasego akams` *also* needs to be able to point at the location where the assets.json was uploaded. This isn't contained inside the assets.json file and can't be guessed. We need to download the publish manifest from microsoft-go to recover this information. So, this PR adds `build-asset-json-publish-manifest` to pass in the path to that file.

The method for creating "floating" names (`go1.23-90bcc55-20240626.1.windows-amd64.zip` -> `go1.22.windows-amd64.zip`) also depended on the "folder" in blob storage in order to process the filename. This was done to make the process sturdier, but it turns out this isn't necessary given what we now know we can expect the filenames to be. The process has been changed so that only the filename is used as input to generate floating names: nothing before the final `/` has any impact.

Example build that mock-publishes the results from the example build in #143:
* https://dev.azure.com/dnceng/internal/_build/results?buildId=2496297&view=logs&j=19992227-62fb-5b50-4e29-3b72bd33eea1&t=2bf309a5-b7e0-54ec-527b-0b1264c53682&l=748
* (The odd version numbers are because this comes from a build of the main, which doesn't have a nice version. The new test cases demonstrate nice versions.)


gdams:
linter failures need fixing

dagood:
Fixed linter issue in https://github.com/microsoft/go-infra/pull/143, but didn't flow nicely into a CI rebuild here. Updated this PR after merging https://github.com/microsoft/go-infra/pull/143.

gdams (approved):

PR #148, opened 2024-07-29
add:create new branch with updated files

mertakman:
This PR introduces an automated workflow to streamline the process of updating Microsoft Go versions for Azure Linux. Whenever a new Microsoft Go release is published, the following actions are automatically triggered:
Updating the specfile with changelog,
Updating the download URL and version in cgmanifest.json,
Updating the signature hash and download URL link in signatures.json.


cmd/releasego/update-azure-linux.go
```diff
@@ -96,19 +107,79 @@ func updateAzureLinux(p subcmd.ParseFunc) error {
...
+
+	for filePath, newContent := range updatedFiles {
+		// Get the file to update
+		file, _, _, err := client.Repositories.GetContents(ctx, owner, repo, filePath, &github.RepositoryContentGetOptions{Ref: updateBranch})
+		if err != nil {
+			return fmt.Errorf("Failed to get file %q: %v", filePath, err)
+		}
+
+		// Update the file
+		_, _, err = client.Repositories.UpdateFile(ctx, owner, repo, filePath, &github.RepositoryContentFileOptions{
```
dagood, line 138: This would make a separate commit per file, right? Could we use https://pkg.go.dev/github.com/google/go-github/github#GitService.CreateTree -> CreateCommit -> UpdateRef instead? (I've done it before, but from C# code.)

cmd/releasego/update-azure-linux.go
```diff
@@ -96,19 +107,79 @@ func updateAzureLinux(p subcmd.ParseFunc) error {
...
+	if err != nil {
+		return fmt.Errorf("Failed to get ref: %v", err)
+	}
+
+	newRef := &github.Reference{
+		Ref:    github.String(updateBranch),
+		Object: &github.GitObject{SHA: ref.Object.SHA},
+	}
+
+	if _, _, err = client.Git.CreateRef(ctx, owner, repo, newRef); err != nil {
```
dagood, line 120: Will this fail if the ref already exists? Maybe we should create a new ref instead? (In case of a rerun.)

cmd/releasego/update-azure-linux.go
```diff
@@ -96,19 +107,79 @@ func updateAzureLinux(p subcmd.ParseFunc) error {
...
+		if err != nil {
+			return fmt.Errorf("Failed to update file %q: %v", filePath, err)
+		}
+	}
+
+	pr, _, err := client.PullRequests.Create(ctx, owner, repo, &github.NewPullRequest{
+		Title: github.String(generatePRTitleFromAssets(assets)),
+		Head:  github.String(updateBranch),
+		Base:  github.String(baseBranch),
+		Body:  github.String("TODO"),
```
dagood, line 153: FWIW: I think the Azure Linux team will be fine with us having custom info here, not necessarily something based on the template.

cmd/releasego/update-azure-linux.go
```diff
@@ -96,19 +107,79 @@ func updateAzureLinux(p subcmd.ParseFunc) error {
...
-	ar := txtar.Archive{
-		Comment: []byte("Bump version to " + assets.GoVersion().Full()),
-		Files: []txtar.File{
-			{Name: cgManifestFilepath, Data: cgManifestBytes},
-			{Name: golangSignaturesFilepath, Data: golangSignaturesFileBytes},
-			{Name: golangSpecFilepath, Data: []byte(golangSpecFileContent)},
-		},
+	ref, _, err := client.Git.GetRef(ctx, owner, repo, baseBranch)
+	if err != nil {
+		return fmt.Errorf("Failed to get ref: %v", err)
```
dagood, line 112: Nit: errors should start with lowercase.

cmd/releasego/update-azure-linux.go
```diff
@@ -96,19 +107,79 @@ func updateAzureLinux(p subcmd.ParseFunc) error {
...
+	for filePath, newContent := range updatedFiles {
+		// Get the file to update
+		file, _, _, err := client.Repositories.GetContents(ctx, owner, repo, filePath, &github.RepositoryContentGetOptions{Ref: updateBranch})
+		if err != nil {
+			return fmt.Errorf("Failed to get file %q: %v", filePath, err)
+		}
+
+		// Update the file
+		_, _, err = client.Repositories.UpdateFile(ctx, owner, repo, filePath, &github.RepositoryContentFileOptions{
+			Message: github.String("Bump version to " + assets.GoVersion().Full()),
```
dagood, line 139: Reuse PR title func?

cmd/releasego/update-azure-linux.go
dagood: Talked about this in sync: letting it fail helps avoid potentially generating more PRs/branches than we want, and we will have a parameter on the release pipeline to let us specify a new name upon rerun if we truly want a new branch/PR. This also preserves history rather than overwrite an existing branch/PR.

cmd/releasego/update-azure-linux.go
```diff
@@ -286,15 +351,38 @@ func updateCGManifest(buildAssets *buildassets.BuildAssets, cgManifestContent []
...
+			other["downloadUrl"] = fmt.Sprintf(
+				"https://github.com/microsoft/go/releases/download/v%s/%s",
 				buildAssets.GoVersion().Full(),
 				path.Base(buildAssets.GoSrcURL),
 			)
+
+			buf := new(bytes.Buffer)
+			encoder := json.NewEncoder(buf)
+			encoder.SetEscapeHTML(false)
+			if err := encoder.Encode(registration); err != nil {
```
dagood, line 381: The idea of this general area is to preserve the order of the original keys, right? (Which is something I think we do need to do--to avoid thrashing with other bots (or opinionated humans).)

I think since `registration` is a map, this line ends up sorting it again anyway.

Maybe we could implement something like https://github.com/golang/go/issues/27179#issuecomment-587528269 to preserve order without quite as much complication?

It could also be simpler to treat the file as only text rather than actually fully parsing it as JSON, so it's easier to leave most of the file untouched no matter what other people do to the file.

mertakman: > The idea of this general area is to preserve the order of the original keys, right? (Which is something I think we do need to do--to avoid thrashing with other bots (or opinionated humans).)
> 
> I think since `registration` is a map, this line ends up sorting it again anyway.
> 
> Maybe we could implement something like [golang/go#27179 (comment)](https://github.com/golang/go/issues/27179#issuecomment-587528269) to preserve order without quite as much complication?
> 
> It could also be simpler to treat the file as only text rather than actually fully parsing it as JSON, so it's easier to leave most of the file untouched no matter what other people do to the file.

So I did the another thing mentioned in same issue:
https://github.com/golang/go/issues/27179#issuecomment-848662008

dagood: My worry here is that a bot will change the key order back to what it was, no matter how minimal the difference is. Then next time we run the update, our bot will change it back to "our" order, and then they'll swap it back, and so on. (And, not necessarily an Azure Linux bot: any other team doing auto-updates too.)

Ultimately this just means things are harder to review/`blame` and the Git repo might get a tiny bit larger, so I suppose it's up to Azure Linux. And, if they really want to avoid this problem, they need to add CI that verifies that this file is laid out in a prescriptive order.

cmd/releasego/update-azure-linux.go
```diff
@@ -96,19 +115,88 @@ func updateAzureLinux(p subcmd.ParseFunc) error {
...
+func generateUpdateBranchNameFromAssets(assets *buildassets.BuildAssets) string {
+	return fmt.Sprintf("refs/heads/update-go-%s", assets.GoVersion().Full())
+}
+
+func generatePRTitleFromAssets(assets *buildassets.BuildAssets) string {
+	return fmt.Sprintf("Bump Go Version to %s", assets.GoVersion().Full())
+}
+
+func GeneratePRDescription(upgradePipelineRunID, buddyBuildID string) string {
+	template := `Bump Go Version to 1.22.5-2
```
gdams, line 192: This needs to be a variable

mertakman: fixed, thanks for mentioning.

gdams:
@mertakman the tests are failing now

mertakman:
> @mertakman the tests are failing now

Fixed

cmd/releasego/update-azure-linux.go
```diff
@@ -305,12 +423,17 @@ func updateCGManifest(buildAssets *buildassets.BuildAssets, cgManifestContent []
...
+	encoder.SetEscapeHTML(false)
+	encoder.SetIndent("", "  ")
+
+	err := encoder.Encode(cgManifest) // Use indentation for readability
 	if err != nil {
 		return nil, fmt.Errorf("failed to marshal updated cgmanifest.json: %w", err)
 	}
 
-	updatedCgManifestContent = append(updatedCgManifestContent, '\n') // Add a newline at the end
+	buf.WriteByte('\n') // Add a newline at the end
```
gdams, line 436: remove this

cmd/releasego/update-azure-linux.go
mertakman: After discussion with azure linux team member we decided to go with squash and merge, this removes a  need for making everything in single commit.

cmd/releasego/update-azure-linux.go
mertakman: Fixed with using strudt instead of map for json marshalling.

gdams (approved):
LGTM although I'll let @dagood give it another review before merging

cmd/releasego/update-azure-linux.go
```diff
@@ -96,19 +116,88 @@ func updateAzureLinux(p subcmd.ParseFunc) error {
...
+func generateUpdateBranchNameFromAssets(assets *buildassets.BuildAssets) string {
+	return fmt.Sprintf("refs/heads/update-go-%s", assets.GoVersion().Full())
+}
+
+func generatePRTitleFromAssets(assets *buildassets.BuildAssets) string {
+	return fmt.Sprintf("Bump Go Version to %s", assets.GoVersion().Full())
+}
+
+func GeneratePRDescription(upgradePipelineRunID, buddyBuildID string, assets *buildassets.BuildAssets) string {
+	template := `Bump Go Version to %s
```
dagood, line 193: `Sprintf` calls this `format`, which I'd prefer--calling it `template` implies text/template to me.

dagood, line 193: `const template =` rather than `:=`, to allow arg mismatches to be detected by `go vet` and the linter.

cmd/releasego/update-azure-linux.go
```diff
@@ -96,19 +116,88 @@ func updateAzureLinux(p subcmd.ParseFunc) error {
...
+
+		// Update the file
+		_, _, err = client.Repositories.UpdateFile(ctx, owner, repo, filePath, &github.RepositoryContentFileOptions{
+			Message: github.String(generatePRTitleFromAssets(assets)),
+			Content: newContent,
+			SHA:     file.SHA,
+			Branch:  github.String(updateBranch),
+		})
+		if err != nil {
+			return fmt.Errorf("Failed to update file %q: %v", filePath, err)
```
dagood, line 154:
```suggestion
			return fmt.Errorf("failed to update file %q: %v", filePath, err)
```

A few more capital errors earlier in the file, too.

cmd/releasego/update-azure-linux.go
```diff
@@ -286,12 +392,11 @@ func updateCGManifest(buildAssets *buildassets.BuildAssets, cgManifestContent []
...
-	for i := range cgManifest.Registrations {
-		registration := &cgManifest.Registrations[i].Component.Other
-		if registration.Name == "golang" {
-			registration.Version = buildAssets.GoVersion().MajorMinorPatch()
-			registration.DownloadURL = fmt.Sprintf(
-				"https://github.com/microsoft/go/releases/download/%s/%s",
+	for i, reg := range cgManifest.Registrations {
+		if reg.Component.Other.Name == "golang" {
+			cgManifest.Registrations[i].Component.Other.Version = buildAssets.GoVersion().MajorMinorPatch()
+			cgManifest.Registrations[i].Component.Other.DownloadURL = fmt.Sprintf(
```
dagood, line 398: Why change away from the pointer?

cmd/releasego/update-azure-linux.go
```diff
@@ -36,14 +37,28 @@ See https://github.com/microsoft/go-lab/issues/79
...
+	var updateBranch string
+	var buddyBuildID string
+	var upgradePipelineRunID string
+
+	flag.StringVar(&owner, "owner", "microsoft", "The owner of the repository.")
+	flag.StringVar(&repo, "repo", "azurelinux", "The repository to update.")
+	flag.StringVar(&baseBranch, "base-branch", "refs/heads/3.0-dev", "The base branch to download files from.")
+	flag.StringVar(&updateBranch, "update-branch", "", "The target branch to update files in.")
+	flag.StringVar(&buildAssetJSON, "build-asset-json", "assets.json", "The path of a build asset JSON file describing the Go build to update to.")
+	flag.StringVar(&buddyBuildID, "buddy-build-id", "", "The job ID for the buddy build in Azure DevOps")
```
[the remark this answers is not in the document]
gdams: Thinking aloud here, does the buddy build need the branch or the PR to exist in order to run? We might be able to create the branch and push the code, run the buddy build and then raise the PR afterwards?

cmd/releasego/update-azure-linux.go
```diff
@@ -96,19 +116,88 @@ func updateAzureLinux(p subcmd.ParseFunc) error {
...
+
+func generatePRTitleFromAssets(assets *buildassets.BuildAssets) string {
+	return fmt.Sprintf("Bump Go Version to %s", assets.GoVersion().Full())
+}
+
+func GeneratePRDescription(upgradePipelineRunID, buddyBuildID string, assets *buildassets.BuildAssets) string {
+	template := `Bump Go Version to %s
+Upgrade pipeline run -> https://dev.azure.com/mariner-org/mariner/_build/results?buildId=%s&view=results
+
+Buddy build -> https://dev.azure.com/mariner-org/mariner/_build/results?buildId=%s&view=results
```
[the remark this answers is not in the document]
mertakman: I started a Teams thread about this to discuss further. Let's see where it evolves and then we can apply the change here.

cmd/releasego/update-azure-linux.go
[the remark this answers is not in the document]
dagood: The PR specifically, currently. In the past, we've been told directly that we need to submit the PR first and then edit in the build ID.

Here's the parameter doc for the pipeline:

https://dev.azure.com/mariner-org/mariner/_build?definitionId=2190

> image

Our branch is in a fork, so (unless there's some secret mechanism--which may be unintentional as far as Azure Linux is concerned) we can't specify our branch directly as a topic branch.

PR #151, opened 2024-08-06
Wait for MAR images to match expected `go version`

dagood:
* Resolves https://github.com/microsoft/go/issues/1090

This is only intended as a broad check. If we're trying to ship 1.21.5-1 and the image contains 1.210.5, then something went seriously wrong, and this would catch it. This PR changes it from a manual step to an automated one.

Test run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2510014&view=results
I tested the polling case manually, locally, with versions that don't exist.

The value of this check is limited: `go version` only outputs 1.21.5, so if we're building 1.21.5-2 that replaces 1.21.5-1, it will potentially succeed too early. But, this limitation has always been there--this PR only automates the check, making it clearer that the potential problem exists.

* Related: https://github.com/microsoft/go/issues/262
* This PR isn't good enough as a way to *generally* make sure MAR has all our images. We actually should set up this .NET Docker feature: https://github.com/microsoft/go/issues/1258

Also fixes the pipeline: it was ignoring `runPublishAnnouncement`.

### DIFF
--- a/cmd/releasego/wait-latest-mar-go-version.go
+++ b/cmd/releasego/wait-latest-mar-go-version.go
@@ -1,0 +1,105 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package main
+
+import (
+	"errors"
+	"flag"
+	"fmt"
+	"log"
+	"os/exec"
+	"strings"
+	"time"
+
+	"github.com/microsoft/go-infra/executil"
+	"github.com/microsoft/go-infra/goversion"
+	"github.com/microsoft/go-infra/subcmd"
+)
+
+const marRepo = "mcr.microsoft.com/oss/go/microsoft/golang"
+
+func init() {
+	subcommands = append(subcommands, subcmd.Option{
+		Name:    "wait-latest-mar-go-version",
+		Summary: "Wait (poll) for the latest Microsoft Go images on MAR to match specified versions.",
+		Description: `
+
+Given a list of Go versions, constructs tag names for each one's major version (1.20.3-1 -> 1.20)
+and repeatedly tries to "docker pull" that tag and then probe the image using
+"docker run ... go version" to see if it contains the expected major+minor+patch Go version.
+`,
+		Handle: waitMarGoVersion,
+	})
+}
+
+func waitMarGoVersion(p subcmd.ParseFunc) error {
+	versionList := flag.String(
+		"versions", "",
+		"[Required] A list of full or partial microsoft/go version numbers (major.minor.patch[-revision[-suffix]]). Separated by commas.")
+
+	timeout := flag.Duration("timeout", time.Minute*5, "Time to wait before giving up.")
+	pollDelay := flag.Duration("poll-delay", time.Second*10, "Time to wait between each poll attempt.")
+
+	if err := p(); err != nil {
+		return err
+	}
+
+	if *versionList == "" {
+		return errors.New("no versions specified")
+	}
+
+	var checkers []func() (bool, error)
+	for _, version := range strings.Split(*versionList, ",") {
+		v := goversion.New(version)
+		tag := marRepo + ":" + v.MajorMinor()
+
+		expect := "go" + v.MajorMinorPatchPrerelease() + " "
+
+		checkers = append(checkers, func() (bool, error) {
+			pullCmd := exec.Command("docker", "pull", tag)
+			if err := executil.Run(pullCmd); err != nil {
+				return false, err
+			}
+			versionCmd := exec.Command("docker", "run", "--rm", tag, "go", "version")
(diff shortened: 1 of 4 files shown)


gdams (approved):

cmd/releasego/wait-latest-mar-go-version.go
```diff
@@ -0,0 +1,105 @@
...
+		Handle: waitMarGoVersion,
+	})
+}
+
+func waitMarGoVersion(p subcmd.ParseFunc) error {
+	versionList := flag.String(
+		"versions", "",
+		"[Required] A list of full or partial microsoft/go version numbers (major.minor.patch[-revision[-suffix]]). Separated by commas.")
+
+	timeout := flag.Duration("timeout", time.Minute*5, "Time to wait before giving up.")
```
mertakman, line 41: I think instead of time.Minute*5 , 5*time.Minute is better approach,[ also how stdlib of Go does this ](https://cs.opensource.google/go/go/+/refs/tags/go1.22.6:src/time/time.go;l=637)

cmd/releasego/wait-latest-mar-go-version.go
```diff
@@ -0,0 +1,105 @@
...
+	pollDelay := flag.Duration("poll-delay", time.Second*10, "Time to wait between each poll attempt.")
```
mertakman, line 42: same here.

cmd/releasego/wait-latest-mar-go-version.go
```diff
@@ -0,0 +1,105 @@
...
+			ok, err := tag()
+			if err != nil {
+				return fmt.Errorf("unexpectedly failed check: %v", err)
+			}
+			if !ok {
+				missing = true
+			}
+		}
+
+		if !missing {
```
mertakman, line 94: Instead of continuing to search after a missing item is found, we can optimize the code by checking a variable called allFound. If allFound becomes false, there's no need to continue the loop since we know not all items are present.

Here's a revised version of the code snippet:
```go
		allFound := true
		for _, check := range checkers {
			ok, err := check()
			if err != nil {
				return fmt.Errorf("unexpectedly failed check: %v", err)
			}
			if !ok {
				allFound = false
				break
			}
		}
		
		if allFound {
			log.Println("All versions found. Done.")
			return nil
		}
```

cmd/releasego/wait-latest-mar-go-version.go
```diff
@@ -0,0 +1,105 @@
...
+		})
+	}
+
+	// Make our logs stand out from Docker's.
+	log.Default().SetPrefix("---- ")
+
+	start := time.Now()
+	end := start.Add(*timeout)
+
+	for {
```
mertakman, line 81: By using for time.Now().Before(end), we can directly integrate the timeout logic into the loop's condition, making the code more concise. This approach cleanly separates the polling logic within the loop's scope, improving readability and maintainability.

```go
	for time.Now().Before(end) {
        // all of current loop logic comes here
        }
        return fmt.Errorf("exceeded timeout (%v) waiting for versions", *timeout)
```

cmd/releasego/wait-latest-mar-go-version.go
```diff
@@ -0,0 +1,105 @@
...
+			log.Printf("Finding %q in %q: %v\n", expect, out, found)
+			return found, nil
+		})
+	}
+
+	// Make our logs stand out from Docker's.
+	log.Default().SetPrefix("---- ")
+
+	start := time.Now()
+	end := start.Add(*timeout)
```
mertakman, line 79: we didn't used start in nowhere other than calculating the end. Lets simplify it into
`go
 end := start.Add(*timeout)
`

mertakman (changes requested):
I've reviewed your code and have a few suggestions for making it cleaner and more precise. These are minor nitpicks, but I believe they'll enhance the overall quality of the code.

cmd/releasego/wait-latest-mar-go-version.go
dagood: Agreed, I think it reads better, more like "5 minutes".

cmd/releasego/wait-latest-mar-go-version.go
dagood: A `break` after `missing = true` would do the same thing as far as I can tell, but I intentionally didn't do this to help with diagnosis: if a human sees this task failed and needs to know what's going on, it's more useful to them to see the full picture.

A `break` would be a normal thing to do here to be sure, so definitely worth adding a comment to explain.

(I do think `check` is a better name than `tag` though. 😄 Not sure why I called it that.)

mertakman (approved):

mertakman:
Looks good to go

PR #155, opened 2024-08-19
fix:add azdo pipeline trigger command

mertakman:


cmd/releasego/azdo-pipeline-trigger.go
```diff
@@ -0,0 +1,14 @@
+package main
+
+import "github.com/microsoft/go-infra/subcmd"
+
+func init() {
+	subcommands = append(subcommands, subcmd.Option{
+		Name:    "azdo-pipeline-trigger",
+		Summary: "Trigger the Azure DevOps pipeline for Azure Linux.",
```
dagood, line 8: I think it should be possible to do this by adding more flags to https://github.com/microsoft/go-infra/blob/main/cmd/releasego/build-pipeline.go that support cross-org auth.

gdams: agreed, I'll close this for now

PR #157, opened 2024-08-23
Flatten deep nesting of release pipeline steps

dagood:
The deep nesting in eng/pipelines/release-build-pipeline.yml has always been a readability issue and makes it a lot harder to add steps to these pipelines than it should be. This PR is an attempt to fix these problems by minimizing the amount of deeply nested AzDO yaml code, allowing for the actual steps to be defined in a reasonable way.

Inspired by the classic eyebrow-raising `for i := 0; i < 10; i++ { switch(i) { ... } }`. 😄 It seems to me that the AzDO yaml lang's limitations make it difficult to do something any better while using multiple `step`s. (`step`s are nice for browsing in the AzDO UI, so there is a reason to keep using them.)

Check out the pipeline file outside of diff view to make the pattern clearer: https://github.com/microsoft/go-infra/blob/dev/dagood/unindent-release-steps/eng/pipelines/release-build-pipeline.yml#L107-L136

I originally started this as part of https://github.com/microsoft/go-lab/blob/main/docs/adr/0003-use-ddfun-for-akams.md, but realized it would be better to do this refactor in its own PR first. This way I can add new steps without modifying the indentation of all the preceding steps. I need to add at least one more new `poll*` parameter to move publish over to go-infra, which would be much more disruptive in the current arrangement.

If this pattern looks reasonable, we should apply it to the other pipelines with `poll*` parameters too.

May be difficult to review. Git doesn't track moving identical chunks between files in a nice way for us. I've double checked that the steps are the same by copying them over once again and seeing if Git shows any differences: https://gist.github.com/dagood/4c00604d3e9e9ca70f1f7ab393bc031d. The diff is in a way, inverted: it shows that I just added a few newlines and slightly clarified and deduped a comment.

Example test builds:
* Test 1.23.1-1 upstream tag polling. 1.23.1-1 doesn't exist yet, so polled until canceled. Shows that the steps are laid out as expected: https://dev.azure.com/dnceng/internal/_build/results?buildId=2522732&view=results
* Test starting at poll3MicrosoftGoBuildID and with only aka.ms update selected (including /dev/ element): https://dev.azure.com/dnceng/internal/_build/results?buildId=2522717&view=results

### DIFF
--- a/eng/pipelines/release-build-pipeline.yml
+++ b/eng/pipelines/release-build-pipeline.yml
@@ -96,268 +96,41 @@ extends:
               releaseIssue: ${{ parameters.releaseIssue }}
               retryInstructionsFlags: '-checkboxes'
               variables:
-                # Set a variable for each polling parameter. If a variable starts off with 'nil', it will be
-                # overridden with a logging command as the pipeline moves on.
+                # Set a variable for each polling parameter. If a variable starts off with 'nil', it
+                # will be overridden with a logging command as the pipeline moves on. Variables are
+                # dynamic (unlike parameters) and the final state of the variables is checked at the
+                # end of the build to provide a starting point for a retry if necessary.
                 poll1MicrosoftGoPRNumber: ${{ parameters.poll1MicrosoftGoPRNumber }}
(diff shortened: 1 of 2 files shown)


karianna (approved):
I dislike pyramids of doom, this is an improvement :-)

PR #161, opened 2024-09-03
fix:add missing comment field to update-azure-linux command

mertakman:
Adds missing comment field to update-azure-linux. Fixes [this](https://github.com/microsoft/go-lab/issues/102)

### DIFF
--- a/cmd/releasego/update-azure-linux.go
+++ b/cmd/releasego/update-azure-linux.go
@@ -364,8 +364,9 @@ type CGManifest struct {
 
 type Registration struct {
 	Component struct {
-		Type  string `json:"type"`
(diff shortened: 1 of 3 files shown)


cmd/releasego/update-azure-linux.go
```diff
@@ -364,8 +364,9 @@ type CGManifest struct {
 
 type Registration struct {
 	Component struct {
-		Type  string `json:"type"`
-		Other struct {
+		Type    string `json:"type"`
+		Comment string `json:"comment,omitempty"`
```
dagood, line 368: Add a test case that exercises it (or add a bit more to existing test)?

PR #166, opened 2024-09-09
releasego: fix release number reset logic

dagood:
* Fixes https://github.com/microsoft/go-lab/issues/114

Move release number reset logic (and some parsing, for cleanliness) to `updateSpecVersion` func, fix it by considering `Patch`, and add tests.

### DIFF
--- a/cmd/releasego/update-azure-linux.go
+++ b/cmd/releasego/update-azure-linux.go
@@ -276,34 +276,20 @@ func updateSpecFile(assets *buildassets.BuildAssets, changelogDate time.Time, sp
 		oldVersion = goversion.New(matches[2])
 	}
 
-	var oldRelease int
+	var oldRelease string
 	if matches := specFileReleaseRegex.FindStringSubmatch(specFileContent); matches == nil {
 		return "", fmt.Errorf("no Release declaration found in spec content")
 	} else {
-		var err error
-		oldRelease, err = strconv.Atoi(matches[2])
-		if err != nil {
-			return "", fmt.Errorf("failed to parse Release number: %w", err)
-		}
+		oldRelease = matches[2]
 	}
 
-	newVersion := assets.GoVersion().MajorMinorPatch()
-	if strings.Contains(newVersion, "$") {
-		return "", fmt.Errorf("new version %q contains unexpected $", newVersion)
+	newVersion, newRelease, err := updateSpecVersion(assets, oldVersion, oldRelease)
+	if err != nil {
+		return "", err
 	}
+
 	specFileContent = specFileVersionRegex.ReplaceAllString(specFileContent, "${1}"+newVersion)
-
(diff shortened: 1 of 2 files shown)


cmd/releasego/update-azure-linux.go
```diff
@@ -276,34 +276,20 @@ func updateSpecFile(assets *buildassets.BuildAssets, changelogDate time.Time, sp
...
+	var oldRelease string
 	if matches := specFileReleaseRegex.FindStringSubmatch(specFileContent); matches == nil {
 		return "", fmt.Errorf("no Release declaration found in spec content")
 	} else {
-		var err error
-		oldRelease, err = strconv.Atoi(matches[2])
-		if err != nil {
-			return "", fmt.Errorf("failed to parse Release number: %w", err)
-		}
+		oldRelease = matches[2]
```
karianna, line 283: not sure of std in this Go code base, but if 2 is a magic (index) number, replace with a const?

cmd/releasego/update-azure-linux.go
```diff
@@ -276,34 +276,20 @@ func updateSpecFile(assets *buildassets.BuildAssets, changelogDate time.Time, sp
...
-	// Azure-Linux-specific fix, so this is independent of Microsoft Go releases. When updating to a
-	// new major/minor version (as semver refers to them), reset release to 1.
-	var newRelease int
-	if assets.GoVersion().MajorMinor() != oldVersion.MajorMinor() {
-		newRelease = 1
-	} else {
-		newRelease = oldRelease + 1
+	newVersion, newRelease, err := updateSpecVersion(assets, oldVersion, oldRelease)
+	if err != nil {
+		return "", err
```
karianna, line 288: I assume the caller will deal with empty string just fine?

karianna (approved):
Tests make me happy.  Feel free to ignore my other comments, I'm more asking Q's because I'm new here.

cmd/releasego/update-azure-linux.go
dagood: It's just the group number in the reg exp, but yeah, that could be stored in a constant adjacent to the exp itself.

cmd/releasego/update-azure-linux.go
dagood: A detailed doc comment for this func might mention "returns the spec file content or an error". The presence of an error determines whether it's meaningful to use the string for anything. Some functions go into more detail about what exactly is returned when it's useful for the func to promise something more specific, but generally it's simply incorrect to depend on the other values that a func returns when it returns an error.

Various idioms come out of that. If there isn't a reason to return in some other way, typically you either:

* `return (something), (something), ..., nil` or
* `return (zero), (zero), ..., err`

In this case, `""` is just the zero string value.

This is a clever proposal to make that idiom easier to write, but unfortunately it didn't make it:

* https://github.com/golang/go/issues/61372

PR #170, opened 2024-09-20
update-azure-linux: add upstream arg for fork workflow, update GitHub client

dagood:
Adds `-upstream` to the `update-azure-linux` command, and uses a fork contribution approach when upstream != owner.

Creates the fork if necessary so we don't need to log into the bot to pre-create the fork. This also makes it so that if we delete the fork to clean up old data, the next run will create it again automatically.

* Example: https://github.com/dagood/azurelinux/pull/8
* By https://dev.azure.com/dnceng/internal/_build/results?buildId=2542636&view=results

This also implements single-commit PRs using the tree API and makes some parts of the command more robust to transitive failures (e.g. rate limit exceeded) by wrapping more with Retry.

It still works with a non-fork approach: https://github.com/dagood/azurelinux/pull/9

I also updated the version of the GitHub library we're using. This adds some APIs that I ended up using, and it's something we should do anyway. (The old library was from 2018.)

**Turn on "ignore whitespace" to review cmd/releasego/update-azure-linux.go with a better diff.**

* Fixes https://github.com/microsoft/go-lab/issues/103

### DIFF
--- a/cmd/releasego/update-azure-linux.go
+++ b/cmd/releasego/update-azure-linux.go
@@ -130,12 +130,12 @@ func updateAzureLinux(p subcmd.ParseFunc) error {
 		// working on, making sure our PR won't revert an unfortunately-timed upstream merge.
 		upstreamRef, _, err := client.Git.GetRef(ctx, upstream, repo, baseBranch)
 		if err != nil {
-			return fmt.Errorf("failed to get ref %v: %v", baseBranch, err)
+			return fmt.Errorf("failed to get ref %v: %w", baseBranch, err)
 		}
 		upstreamCommitSHA := upstreamRef.Object.GetSHA()
 		upstreamCommit, _, err := client.Git.GetCommit(ctx, upstream, repo, upstreamCommitSHA)
 		if err != nil {
-			return fmt.Errorf("failed to get commit %v: %v", upstreamCommitSHA, err)
+			return fmt.Errorf("failed to get commit %v: %w", upstreamCommitSHA, err)
 		}
 
 		golangSpecFileBytes, err := downloadFileFromRepo(ctx, client, upstream, repo, upstreamCommitSHA, golangSpecFilepath)
@@ -217,7 +217,7 @@ func updateAzureLinux(p subcmd.ParseFunc) error {
 			Object: &github.GitObject{SHA: createCommit.SHA},
 		}
 		if _, _, err = client.Git.CreateRef(ctx, owner, repo, newRef); err != nil {
-			return fmt.Errorf("failed to create ref: %v", err)
+			return fmt.Errorf("failed to create ref: %w", err)
 		}
 		// Now that we've created the ref, we can't retry: the name is taken.
 		// This retry loop is over.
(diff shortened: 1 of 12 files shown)


cmd/releasego/update-azure-linux.go
```diff
@@ -61,118 +80,190 @@ func updateAzureLinux(p subcmd.ParseFunc) error {
...
+	if upstream != owner {
+		// Submitting PR via fork. Try to make sure it'll work.
+		if !slices.Contains(patScopes, "workflow") {
+			return fmt.Errorf("the PAT must have 'workflow' scope, but not found in list: %v", patScopes)
+		}
 
-	prevGoArchiveName, err := extractGoArchiveNameFromSpecFile(golangSpecFileContent)
-	if err != nil {
-		return err
+		// Check if fork already exists, or create it.
```
qmuntal, line 101: nit: This comment confused me. `FetchRepositoryOrNil` doesn't create the fork if it doesn't exist. It is done some lines below by `FullyCreateFork`. Consider adding a blank line below the comment so it is more clear that it applies to the subsequent lines, or split the comment in two: `Check if fork already exists` and `the fork doesn't exist, create it` just before `FullyCreateFork`.

githubutil/githubutil.go
```diff
@@ -156,3 +156,92 @@ func DownloadFile(ctx context.Context, client *github.Client, owner, repo, branc
...
+		return nil
+	}); err != nil {
+		return nil, err
+	}
+
+	return scopes, nil
+}
+
+// FetchRepositoryOrNil fetches a repository from GitHub, or returns nil if it doesn't exist.
+// Returning nil simplifies error handling for the caller if it's expected that the repository may
```
qmuntal, line 215: One the other hand, it complicates the error handling if the repo is expected to exists. Callers will have to check for two conditions: `err != nil && repo != nil`, and they better don't forget the latter, else they will have a nil pointer exception down the line.

The idiomatic and less error-prone way to do this is defining an error sentinel value that is returned only when the repo doesn't exist, see for example [os.ErrNotExist](https://pkg.go.dev/os#ErrNotExist) and [sql.ErrNoRows](https://pkg.go.dev/database/sql#ErrNoRows). This way callers can rely just on the error to implement error-handling logic.

qmuntal (changes requested):

githubutil/githubutil.go
dagood: I expected that callers expecting the repo to exist will simply use `client.Repositories.Get`, which does err out. 😄 But yeah, on second thought, there's no reason not to code this as a more general-purpose wrapper. (I wrote a very similar comment for the `DownloadFile` implementation originally, although it was already being used as a general-purpose wrapper.)

cmd/releasego/update-azure-linux.go
```diff
@@ -61,118 +80,190 @@ func updateAzureLinux(p subcmd.ParseFunc) error {
...
-		return err
-	}
+	// If anything fails here, retry from the beginning to use a fresh base commit.
+	// Some individual steps also have their own retries; this is fine.
+	if err := githubutil.Retry(func() error {
+		// Find details about the state of the upstream base branch. This pins the commit that we're
+		// working on, making sure our PR won't revert an unfortunately-timed upstream merge.
+		upstreamRef, _, err := client.Git.GetRef(ctx, upstream, repo, baseBranch)
+		if err != nil {
+			return fmt.Errorf("failed to get ref %v: %v", baseBranch, err)
```
dagood, line 131: `%v` rather than `%w` is actually bad here (and a few other places) because `Retry` needs to handle `*github.RateLimitError`.

qmuntal (approved):

karianna (approved):

PR #182, opened 2024-10-10
Add eng/install-scripts/microsoft-go-install.ps1

dagood:
* For https://github.com/microsoft/go/issues/1324
* In the short term, this is how we could get Microsoft Go in our own pipelines and keep it up to date. https://github.com/microsoft/go-lab/issues/109

I started with the [.NET install script](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-install-script) and turned it into a Microsoft Go install script while generally keeping features/scenarios and making it work with tar.gz archives. I kept my changes in a separate commit.

From here, we can decide what we actually want to maintain and perhaps cut a lot.

A feature I didn't include is installing the "correct" Go based on the project. In .NET a `global.json` may specify a .NET version, in Go we could leverage the `go.mod`'s `toolchain go1.23.2` directive. However, there's no room for our revision `-1`/`-2` here, so it wouldn't be totally reproducible. (It would have to download the newest revision we release of the matching 1.x.y.)

I ran this with these (all amd64):

* Windows:
  * PowerShell (most restrictive)
  * PowerShell Core 
* Linux (Fedora)
  * PowerShell Core


.github/workflows/install-script-pwsh-test.yml
```diff
@@ -0,0 +1,38 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+# This job tests that the install script in PowerShell works as expected when
+# actually downloading Microsoft Go, not only the unit tests.
```
karianna, line 5:
```suggestion
# actually downloading Microsoft Go, not only just running the unit tests.
```

.github/workflows/install-script-pwsh-test.yml
```diff
@@ -0,0 +1,38 @@
...
+jobs:
+  test:
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - windows-latest
+          - windows-2019
+          - ubuntu-latest
+          - ubuntu-20.04
```
karianna, line 24: Not 22.04?  I assume latest would jump to 24.04 and so on?

karianna: Same applies to Windows above

install/powershellscript/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+### Set up a CI test to ensure the script is up to date
+
+First, make sure dependabot is working.
+It will submit PRs that update the microsoft/go-infra dependency automatically.
+
+Then, add a CI test that runs the following commands in the directory where the script is stored:
+
+```
+go install github.com/microsoft/go-infra/install/powershellscript/cmd/microsoftgoinstallscript
```
qmuntal, line 99: Can we omit this command? We already explained how to install `microsoftgoinstallscript` in a previous section, and it should be necessary to reinstall it every time a command is executed.

install/powershellscript/README.md
```diff
@@ -0,0 +1,112 @@
...
+microsoftgoinstallscript -check
+```
+
+This test will fail if the script is not up to date, and you can update the branch using these commands in the script directory:
+
+```
+go install github.com/microsoft/go-infra/install/powershellscript/cmd/microsoftgoinstallscript
```
qmuntal, line 106: Same as previous comment.

install/powershellscript/microsoft-go-install.ps1
```diff
@@ -0,0 +1,898 @@
...
+    - The toolset needs to be installed without user interaction and without admin rights.
+    - The toolset installation doesn't need to persist across multiple CI runs.
+    Visit https://github.com/microsoft/go for a list of other ways to install Microsoft Go.
+
+.PARAMETER Version
+    Default: Latest
+    Download the specified version. Supports some aliases. Possible values:
+    - Latest - the most recent major version.
+    - Previous - the second most recent major version.
+    - 2-part version in format go1.A - represents a specific major version.
```
qmuntal, line 20: We should document whether prerelease versions are supported or not, i.e. `1.23rc1`.

qmuntal (changes requested):

qmuntal (commented):
Meta question: can we omit the `microsoft` prefix from the script name and from the new package name? In the future we might want to support downloading the upstream toolchain using this same script and just changing a parameter (which will always default to the Microsoft toolchain). If that happens, then having the Microsoft prefix will feel strange.

I guess the Microsoft prefix is there to avoid conflicting with upstream tooling, but I don't think it will ever conflict. Google want provide a powershell script, and even if it did, users installing both could rename our script with whatever name they like.

qmuntal (commented):
Meta comment: I don't expect that we update the PowerShell script nor the `microsoftgoinstallscript` package often, maybe one or two times per year, if any. On the other hand, the `go-infra` is updated quite more often. Also, `microsoftgoinstallscript` doesn't depend on other `go-infra` packages.

With this in mind, I would suggest planting a `go.mod` file in `install/powershellscript/cmd/microsoftgoinstallscript` so that package gets promoted to being it's own module. Then we can version it independently of 'go-infra'. Well, we are currently versioning `go-infra`, but I think we should definitely version `microsoftgoinstallscript`.

install/powershellscript/README.md
```diff
@@ -0,0 +1,112 @@
...
+```
+microsoftgoinstallscript
+```
+
+### Updating the script using `microsoftgoinstallscript`
+
+To update the script, run this in the directory where the script is stored:
+
+```
+go get github.com/microsoft/go-infra/install/powershellscript/cmd/microsoftgoinstallscript@latest
```
qmuntal, line 85: The package name is too verbose for my old eyes. Users will mainly want to install the CLI rather than using the  small Go API in `github.com/microsoft/go-infra/install/powershellscript`. When the principal "product" is a CLI and not the Go API, it is common to not follow the usual `pkgname/cmd/cliname` convention but use `cliname/pkgname` instead. This way the CLI is easier to install (requires less typing!).

Also, the `microsoftgoinstallscript` cli can potentially be made more generic and install other script types. Even if we don't eventually do it, I don't see a lot of value on putting it under a folder named `powershellscript`, as we are artificially limiting the scope of the cli.

With all this in mind, this is my suggestion:

- Rename `github.com/microsoft/go-infra/install/powershellscript` to `github.com/microsoft/go-infra/goinstallscript`.
- Move all files in `github.com/microsoft/go-infra/goinstallscript` to `github.com/microsoft/go-infra/goinstallscript/powershell`.
- Move `github.com/microsoft/go-infra/goinstallscript/cmd/microsoftgoinstallscript/main.go` to `github.com/microsoft/go-infra/goinstallscript`.

If we do this, then installing the cli will look like this:

```
go get github.com/microsoft/go-infra/goinstallscript@latest
```

qmuntal (commented):
Meta question: can we omit the microsoft prefix from the script name and from the new package name? In the future we might want to support downloading the upstream toolchain using this same script and just changing a parameter (which will always default to the Microsoft toolchain). If that happens, then having the Microsoft prefix will feel strange.

I guess the Microsoft prefix is there to avoid conflicting with upstream tooling, but I don't think it will ever conflict. Google want provide a powershell script, and even if it did, users installing both could rename our script with whatever name they like.

install/powershellscript/cmd/microsoftgoinstallscript/main.go
```diff
@@ -0,0 +1,59 @@
+// Copyright (c) Microsoft Corporation.
```
qmuntal, line 1: Meta question: can we omit the microsoft prefix from the script name and from the new package name? In the future we might want to support downloading the upstream toolchain using this same script and just changing a parameter (which will always default to the Microsoft toolchain). If that happens, then having the Microsoft prefix will feel strange.

I guess the Microsoft prefix is there to avoid conflicting with upstream tooling, but I don't think it will ever conflict. Google want provide a powershell script, and even if it did, users installing both could rename our script with whatever name they like.

.github/workflows/install-script-pwsh-test.yml
```diff
@@ -0,0 +1,38 @@
+# Copyright (c) Microsoft Corporation.
```
qmuntal, line 1: Meta comment: I don't expect that we update the PowerShell script nor the microsoftgoinstallscript package often, maybe one or two times per year, if any. On the other hand, the go-infra is updated quite more often. Also, microsoftgoinstallscript doesn't depend on other go-infra packages.

With this in mind, I would suggest planting a go.mod file in install/powershellscript so that package gets promoted to being it's own module. Then we can version it independently of 'go-infra'. Well, we are currently versioning go-infra, but I think we should definitely version microsoftgoinstallscript.

dagood: Huh, I've always assumed those still get versioned with the surrounding module, but finally saw https://go.dev/doc/modules/managing-source#multiple-module-source for how that works.

Yeah, we would produce a lot of useless dependabot update PRs for people if we didn't version it separately.

karianna (approved):

goinstallscript/main.go
```diff
@@ -0,0 +1,90 @@
...
+	"bytes"
+	"flag"
+	"fmt"
+	"os"
+	"os/exec"
+
+	"github.com/microsoft/go-infra/goinstallscript/powershell"
+)
+
+// scriptEmitPackage is a Go package that emits the PowerShell script content. The goinstallscript
```
qmuntal, line 16: I'm afraid I don't get the use-case that this indirection tries to tackle. Why someone would want to install multiple versions of the script? How other steps would be necessary to accomplish that?

Also, this new `go run` approach precludes running `goinstallscript` in environments without `go` installed. I think it was a nice feature that before you could plant `goinstallscript` in a brand new environament and just execute it to install the powershell script. Note that we could support both use-cases by falling back to `powershell.Content` if [exec.LookPath("go")](https://pkg.go.dev/os/exec#LookPath).

dagood: The main scenario here is:

1. Dev installs github.com/microsoft/go-infra/goinstallscript v0.0.1.
2. Dev sets up their repo with v0.0.1 of the script.
3. Some time passes.
4. We update the script, dependabot submits a PR to v0.0.2 of github.com/microsoft/go-infra/goinstallscript.
5. CI fails because the v0.0.1 content is still in the repo.
6. Dev gets the branch and runs `goinstallscript`...

* Without indirection, `goinstallscript` is still `v0.0.1`, so it puts the `v0.0.1` content in the file again, accomplishing nothing.
* With indirection, they get the `v0.0.2` content without having to reinstall `goinstallscript`. They're able to push the change back up to the repo and move on.

Some other similar mixups could happen if the dev works on multiple repos, e.g. one uses `v0.0.2` of the script and another uses `v0.0.3`--relying on the user-wide `goinstallscript`'s embedded content means they need to run `go install` very often.

This is all about https://github.com/microsoft/go-infra/pull/182#discussion_r1806762461.

But... now that I'm faced with this again, I'm realizing that we can go back to the direct package and simply document that people should use `go run`, not `go install`. I didn't go back to reevaluate once I realized `go run` is useful here. 😄 

---

FWIW, so far, I've only considered the scenario of using the *script* to install Go without Go, and `goinstallscript` is purely for someone who's trying to set up CI in a Go repo they've been working on (so they would already have Go). I'm not able to think of a scenario where it's reasonable to transfer a prebuilt `goinstallscript` binary to a machine and not ok to transfer `go-install.ps1` itself, though...

gdams (approved):

goinstallscript/main.go
dagood: @qmuntal, does the use case and simplification make sense?

qmuntal (approved):

PR #183, opened 2024-10-23
add conditional flag to disable automerge for sync PRs

gdams:
We're investigating switching to GitHub workflows to handle this so we can restrict the token permissions on our GitHub bots

### DIFF
--- a/eng/pipelines/upstream-sync-pipeline.yml
+++ b/eng/pipelines/upstream-sync-pipeline.yml
@@ -59,5 +59,6 @@ extends:
                     -github-user microsoft-golang-bot `
                     -github-pat $(BotAccount-microsoft-golang-bot-PAT) `
                     -github-pat-reviewer $(BotAccount-microsoft-golang-review-bot-PAT) `
+                    -auto-merge false `
                     -azdo-dnceng-pat $(dn-bot-dnceng-build-rw-code-rw)
                 displayName: Sync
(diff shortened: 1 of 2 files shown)


eng/pipelines/upstream-sync-pipeline.yml
```diff
@@ -59,5 +59,6 @@ extends:
                     -github-user microsoft-golang-bot `
                     -github-pat $(BotAccount-microsoft-golang-bot-PAT) `
                     -github-pat-reviewer $(BotAccount-microsoft-golang-review-bot-PAT) `
+                    -auto-merge false `
```
dagood, line 62: We should be sure this works before hard-coding it and potentially blocking ongoing auto-sync. It should work fine to push this change internally and run the pipeline on the dev branch to see what happens.

dagood (changes requested):

eng/pipelines/upstream-sync-pipeline.yml
gdams: I've triggered https://dev.azure.com/dnceng/internal/_build/results?buildId=2567448&view=logs&s=ddda3e96-b7f2-5f09-e4e5-8b5fbdbd8459 to test

gdams:
Screenshot 2024-10-23 at 19 24 17

Not sure this is going to work as we don't seem to be able to give the Msft org permission to allow actions to approve PRs (even with elevated access)

dagood:
Closing, test complete.

PR #195, opened 2024-12-03
Increase release-build timeouts

dagood:
* microsoft/go PRs are almost exactly 60 minutes now on the Windows jobs. Expand the 60 minute PR poll timeout to 90 minutes.
* The internal build has taken more than 120 minutes a few times recently. Expand to 180 minute timeout.

These timeouts caused some unnecessary retry runs and delays in the recent releases.

### DIFF
--- a/eng/pipelines/steps/release-build-steps.yml
+++ b/eng/pipelines/steps/release-build-steps.yml
@@ -46,7 +46,7 @@ steps:
           -pat '$(GitHubPAT)' \
           -set-azdo-variable poll2MicrosoftGoCommitHash
       displayName: ⌚ Get sync PR merged commit hash
-      timeoutInMinutes: 60
+      timeoutInMinutes: 90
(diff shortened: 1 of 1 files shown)


karianna (approved):

karianna:
I'm happy with this workaround but do we have a separate issue tracking build speed?

qmuntal (approved):

dagood:
> I'm happy with this workaround but do we have a separate issue tracking build speed?

* Closest match: https://github.com/microsoft/go/issues/656
  * Public CI innerloop is essentially the same as internal innerloop.
  * I generally expect these jobs to get slower as more upstream tests are added all the time, and we need to be careful to account for that before making any real over-time conclusions.
* We'll be in a better position to monitor once we have https://github.com/microsoft/go-lab/issues/122. AzDO e.g. obscures agent reservation times, but we can gather fuller data from AzDO APIs once we have a more centralized organizer.
* Publishing (during internal build) is known to be way slower after we moved to Download Center/CDN.

PR #212, opened 2025-02-07
Implement json2junit

qmuntal:
This PR implements a simpler converted from Go JSON output to JUnit XML files.

The implementation is less that 300 lines. I don't expect it to grow in the future, it is feature complete<sup>tm</sup>. The remaining additions in this PR are just test files.

The alternatives are:
- [gotest.tools/gotestsum](https://pkg.go.dev/gotest.tools/gotestsum), which I don't like because it is a test runner, not a text parser, which imposes unnecessary restrictions to our build system.
- [github.com/jstemmer/go-junit-report/v2](https://github.com/jstemmer/go-junit-report), which have not been updated for a while and has an issue that we are hitting in the Microsoft Go CI: https://github.com/jstemmer/go-junit-report/issues/182.

All in all, I think it worth having some tailored parser to cover our limited use-case.


gdams (approved):

dagood (approved):
Nice. Overall, I wonder if it would be useful to also expose it as a library? Then it could be used directly in `_util` without an extra pipeline step (and extra file to manage explicitly). (Then it could also potentially be streamed.)

cmd/json2junit/main.go
```diff
@@ -0,0 +1,289 @@
...
+func run(in, out string) (err error) {
+	entries, err := parseJson(in)
+	if err != nil {
+		return fmt.Errorf("failed to parse JSON: %v", err)
+	}
+	junit, err := convertJsonToJUnit(entries)
+	if err != nil {
+		return fmt.Errorf("failed to convert to JUnit: %v", err)
+	}
+	f, err := os.Create(out)
```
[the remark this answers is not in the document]
qmuntal: Done.

cmd/json2junit/main.go
```diff
@@ -0,0 +1,289 @@
...
+					suite.SystemOut = new(systemOut)
+				}
+				suite.SystemOut.Content += out + "\n"
+				break
+			}
+			testCase := suite.testCase(entry.Test)
+			if testCase == nil {
+				return nil, fmt.Errorf("no run entry for %v", entry.Test)
+			}
+			testCase.systemOut += out + "\n"
```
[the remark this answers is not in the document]
qmuntal: I've switched to a slice of bytes to avoid that many relocations. Thanks.

json2junit/testdata/TestRun/chatty/chatty.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<testsuites>
+<testsuite name="cmd/go" tests="2" failures="0" skipped="0" time="0.011" timestamp="2025-02-07T14:43:16+01:00">
```
dagood, line 3: It might be worth adding a tiny happy path test json that's easy for a person to read in full (without truncation by default) (maybe also easier to debug into). `chatty.json` is the smallest one, but it's still intentionally not short.

Here's one: https://github.com/microsoft/go-infra/commit/828f05cbf4600e119698ed3e3fc3337895935b69

json2junit/testdata/inputs/chatty.json
```diff
@@ -0,0 +1,100 @@
+{"Time":"2025-02-07T14:43:16.4746636+01:00","Action":"start","Package":"cmd/go"}
+{"Time":"2025-02-07T14:43:16.4751944+01:00","Action":"run","Package":"cmd/go","Test":"TestScript"}
```
dagood, line 2: `.jsonl` will make VS Code's syntax checker happy, by the way. ("JSON Lines".)

json2junit/json2junit.go
```diff
@@ -0,0 +1,315 @@
...
+	Skipped   int        `xml:"skipped,attr"`
+	Time      float64    `xml:"time,attr"`
+	Timestamp string     `xml:"timestamp,attr"`
+	SystemOut *systemOut `xml:",omitempty"`
+	TestCases []*junitTestCase
+}
+
+func (s junitTestSuite) testCase(name string) *junitTestCase {
+	for _, tc := range s.TestCases {
+		if tc.Name == name {
```
dagood, line 150: Hmm, it seems like `testCase` and `suite` are called in a decently tight loop, and I wonder if maps are actually still worth the trouble...

json2junit/json2junit.go
```diff
@@ -0,0 +1,315 @@
...
+		}
+		c.suites = append(c.suites, suite)
+	case "run":
+		// start a new test case.
+		_, suite := c.suite(entry.Package)
+		if suite == nil {
+			return fmt.Errorf("no start entry for %v", entry.Package)
+		}
+		testCase := suite.testCase(entry.Test)
+		if testCase == nil {
```
dagood, line 207: Why no error if there's a duplicate test start, like the error for duplicate suite start?

json2junit/json2junit.go
```diff
@@ -0,0 +1,315 @@
...
+	case "pass", "skip", "fail":
+		// finish the current test case or suite.
+		i, suite := c.suite(entry.Package)
+		if suite == nil {
+			return fmt.Errorf("no start entry for %v", entry.Package)
+		}
+		if entry.Test == "" {
+			suite.Time = entry.Elapsed
+			if entry.Action != "fail" {
+				suite.SystemOut = nil
```
dagood, line 249: This is the first time a SystemOut is cleared out, so I think it deserves a comment too.

But that raises a question... should this really be cleared out upon skip, too? (Or can a suite not be skipped? Not sure if I have the terminology lined up in my head.)

```suggestion
				// In case of success, we don't care about the output.
				suite.SystemOut = nil
```

dagood (approved):
Some suggestions, but still looking good.

json2junit/testdata/TestRun/chatty/chatty.xml
qmuntal: Good point. I've replaced all the lengthy samples with shorter and focused ones.

json2junit/testdata/inputs/chatty.json
qmuntal: Didn't know this trick, thanks!

json2junit/json2junit.go
qmuntal: Done

json2junit/json2junit.go
qmuntal: Done 👍

PR #217, opened 2025-02-14
Add GitHub app based authentication methods

gdams:
See https://github.com/gdams/go/pull/10 as an example of how this looks. This will allow us to no longer depend on the microsoft github bot for raising submodule updates. Note that the reviewer bot still uses PAT for now.

The app we're using can be seen here: https://github.com/apps/bot-for-go

I've already requested installation in the Microsoft org via https://github.com/microsoft/github-operations/issues/652

For now I'm leaving the review bot as it is. We could either create a second app for auto approving PRs or investigate branch rules which would allow the app to be exempt from the "required reviews" rule.

* For https://github.com/microsoft/go-lab/issues/152


dagood:
> microsoft-build-of-go-bot

I don't think this name reads well--we didn't build "Go Bot". 😄 A few other ideas:

* Microsoft Bot for Go
* Microsoft build of Go
  * I prefer lowercase `build` for consistency in general.
  * No need to put bot in the name. (It actually looks bad in the PR to have bot at the end--it's already labeled.)
* Bot for Go
  * Presumably this will show up as a Microsoft managed app, so can simplify.
* Build manager for Go 

Putting this out there ahead of any review in case it takes effort to change the name.

gdams:
> > microsoft-build-of-go-bot
> 
> I don't think this name reads well--we didn't build "Go Bot". 😄 A few other ideas:
> 
> * Microsoft Bot for Go
> * Microsoft build of Go
>   
>   * I prefer lowercase `build` for consistency in general.
>   * No need to put bot in the name. (It actually looks bad in the PR to have bot at the end--it's already labeled.)
> * Bot for Go
>   
>   * Presumably this will show up as a Microsoft managed app, so can simplify.
> * Build manager for Go
> * Fork maintenance for Go
> 
> Putting this out there ahead of any review in case it takes effort to change the name.

agreed, I've renamed it to Bot for Go

gdams:
I've added further app support, we can now create github release issues with the app-based auth too. See https://github.com/gdams/go/issues/11

gdams:
The app has been installed in the Microsoft org. I've kicked off a go-infra-upstream-sync pipeline here to test with the app: https://dev.azure.com/dnceng/internal/_build/results?buildId=2645495&view=logs&s=ddda3e96-b7f2-5f09-e4e5-8b5fbdbd8459

gdams:
> The app has been installed in the Microsoft org. I've kicked off a go-infra-upstream-sync pipeline here to test with the app: https://dev.azure.com/dnceng/internal/_build/results?buildId=2645836&view=results

The job failed although it did successfully raise https://github.com/microsoft/go-images/pull/372. @dagood any idea what went wrong?

gdams:
> The app has been installed in the Microsoft org. I've kicked off a go-infra-upstream-sync pipeline here to test with the app: https://dev.azure.com/dnceng/internal/_build/results?buildId=2645928&view=results

That's created https://github.com/microsoft/go-images/pull/374 which looks good to me

githubutil/githubutil.go
```diff
@@ -35,11 +42,116 @@ func NewClient(ctx context.Context, pat string) (*github.Client, error) {
...
+
+	req, err := http.NewRequest("POST", url, nil)
+	if err != nil {
+		return "", err
+	}
+	req.Header.Set("Authorization", "Bearer "+jwt)
+	req.Header.Set("Accept", "application/vnd.github+json")
+
+	client := &http.Client{}
+	resp, err := client.Do(req)
```
qmuntal, line 117: No need to instantiate a new client, can use the default one.

```suggestion
	resp, err := http.DefaultClient.Do(req)
```

sync/sync.go
```diff
@@ -632,7 +677,8 @@ func MakeBranchPRs(f *Flags, dir string, entry *ConfigEntry) ([]SyncResult, erro
 
 		// If we still have unmerged files, 'git commit' will exit non-zero, causing the script to
 		// exit. This prevents the script from pushing a bad merge.
-		if err := run(newGitCmd("commit", "-m", commitMessage)); err != nil {
+		commitArgs := []string{"commit", "-m", commitMessage}
+		if err := run(newGitCmd(commitArgs...)); err != nil {
```
qmuntal, line 681: Is this change necessary?

qmuntal (changes requested):
Looks good! Just some minor inline comments and one meta-comment: should we be using a client ID instead of an App ID? It seems to be the recommended approach: https://github.blog/changelog/2024-05-01-github-apps-can-now-use-the-client-id-to-fetch-installation-tokens/.

sync/sync.go
gdams: nope, reverting

gdams:
@dagood I've removed most of the pipelines changes for now so that we can approve/merge this PR safely and incrementally test the pipelines changes. Right now I'm updating the sync pipeline and the update images pipeline.

qmuntal:
@gdams I think you missed this comment: https://github.com/microsoft/go-infra/pull/217#pullrequestreview-2626818441.

gdams:
> @gdams I think you missed this comment: [#217 (review)](https://github.com/microsoft/go-infra/pull/217#pullrequestreview-2626818441).

Good spot, checking this now

qmuntal (approved):

eng/pipelines/update-images-pipeline.yml
```diff
@@ -73,8 +73,10 @@ extends:
 
               - pwsh: |
                   go run ./cmd/dockerupdatepr `
```
gdams, line 75: @dagood is this pipeline actually being used anymore? When I checked here it hasn't been run in some time https://dev.azure.com/dnceng/internal/_build?definitionId=1040&view=runs

cmd/releasego/get-images-commit.go
```diff
@@ -36,6 +37,9 @@ func handleGetImagesCommit(p subcmd.ParseFunc) error {
 	azdoVarName := flag.String("set-azdo-variable", "", "An AzDO variable name to set to the commit hash using a logging command.")
 	keepTemp := flag.Bool("w", false, "Keep the temporary repository used for polling, rather than cleaning it up.")
 	pollDelaySeconds := flag.Int("poll-delay", 5, "Number of seconds to wait between each poll attempt.")
+	ghClientId := githubutil.BindClientIDFlag()
```
gdams, line 40: @dagood I was actually unsure about this file. As far as I can tell it's just running a `git fetch` and a `git show` command.

Currently we inject the GH PAT into the url like this:

```
-repo 'https://_:$(GitHubPAT)@github.com/$(TargetGoImagesGitHubRepo)' \
```

but I'm not sure it's actually necessary to pass the token to this command as the go-images repo is public right?

gitcmd/gitcmd.go
```diff
@@ -57,11 +71,58 @@ func (a GitHubPATAuther) InsertAuth(url string) string {
...
+		return fmt.Sprintf("https://x-access-token:%v@github.com/%v", token, after)
+	}
+	return url
+}
+
+func (a GitHubAppAuther) InsertHTTPAuth(req *http.Request) {
+	token, err := a.getInstallationToken()
+	if err != nil {
+		log.Printf("Failed to get GitHub App installation token: %v", err)
+		return
```
dagood, line 97: This deserves an `error` return--the caller shouldn't attempt the request. It's not that `req` doesn't need auth, rather there's something wrong with the `GitHubAppAuther` itself. If the caller carries out `req` and gets an error, it'll be obscure, likely a 404, and this log message may not be clearly visible at that point.

gitcmd/gitcmd.go
```diff
@@ -57,11 +71,58 @@ func (a GitHubPATAuther) InsertAuth(url string) string {
...
+	if err != nil {
+		return "", err
+	}
+
+	// Exchange JWT for an installation token
+	token, err := githubutil.FetchInstallationToken(jwt, a.InstallationID)
+	if err != nil {
+		return "", err
+	}
+	return token, nil
```
dagood, line 114: Can/should we be using the same token for more than one request? Is it just difficult to deal with expiration so we're making one per request for now? Worth a followup issue if so, IMO.

githubutil/githubutil.go
```diff
@@ -13,10 +17,13 @@ import (
 	"strings"
 	"time"
 
+	"github.com/golang-jwt/jwt/v5"
 	"github.com/google/go-github/v65/github"
 	"golang.org/x/oauth2"
 )
 
+const githubAPI = "https://api.github.com"
```
dagood, line 25: `githubAPI` isn't all that descriptive. This is also only used in one place:

```go
	url := fmt.Sprintf("%s/app/installations/%d/access_tokens", githubAPI, installationID)
```

Inlining it arguably makes it clearer what the URL will be in context:

```go
	url := fmt.Sprintf("https://api.github.com/app/installations/%d/access_tokens", installationID)
```

gitcmd/gitcmd.go
```diff
@@ -57,11 +71,58 @@ func (a GitHubPATAuther) InsertAuth(url string) string {
...
+	if err != nil {
+		log.Printf("Failed to get GitHub App installation token: %v", err)
+		return
+	}
+	req.Header.Set("Authorization", "token "+token)
+}
+
+func (a GitHubAppAuther) getInstallationToken() (string, error) {
+	// Generate a JWT using the private key
+	jwt, err := githubutil.GenerateJWT(a.ClientID, a.PrivateKey)
```
dagood, line 104: Use `NewInstallationClient` (or a part of it) to reuse code and potentially get the user a better error message?

githubutil/githubutil.go
```diff
@@ -35,11 +42,115 @@ func NewClient(ctx context.Context, pat string) (*github.Client, error) {
...
+	tokenClient := oauth2.NewClient(ctx, tokenSource)
+
+	return github.NewClient(tokenClient), nil
+}
+
+// GenerateJWT generates a JWT for a GitHub App.
+func GenerateJWT(clientID string, privateKey string) (string, error) {
+	privkey, err := base64.StdEncoding.DecodeString(privateKey)
+	if err != nil {
+		return "", fmt.Errorf("failed to decode private key: %v", err)
```
dagood, line 79:
```suggestion
		return "", fmt.Errorf("failed to base64-decode private key: %v", err)
```

githubutil/githubutil.go
```diff
@@ -35,11 +42,115 @@ func NewClient(ctx context.Context, pat string) (*github.Client, error) {
...
+
+	// Create a client using the installation token
+	tokenSource := oauth2.StaticTokenSource(&oauth2.Token{AccessToken: token})
+	tokenClient := oauth2.NewClient(ctx, tokenSource)
+
+	return github.NewClient(tokenClient), nil
+}
+
+// GenerateJWT generates a JWT for a GitHub App.
+func GenerateJWT(clientID string, privateKey string) (string, error) {
```
dagood, line 76:
```suggestion
func GenerateJWT(clientID, privateKey string) (string, error) {
```

githubutil/githubutil.go
```diff
@@ -35,11 +42,115 @@ func NewClient(ctx context.Context, pat string) (*github.Client, error) {
...
+	if err != nil {
+		return "", fmt.Errorf("failed to parse RSA private key: %v", err)
+	}
+
+	now := time.Now().Unix()
+	claims := jwt.MapClaims{
+		"iat": now,       // Issued at time
+		"exp": now + 600, // Expiration time (10 min)
+		"iss": clientID,  // GitHub Client ID
+	}
```
dagood, line 96: Use `jwt.RegisteredClaims` for clarity and avoid doing our own Unix time calculations?

githubutil/githubutil.go
```diff
@@ -35,11 +42,115 @@ func NewClient(ctx context.Context, pat string) (*github.Client, error) {
...
+	if err != nil {
+		return "", fmt.Errorf("failed to sign JWT: %v", err)
+	}
+	return signedToken, nil
+}
+
+func FetchInstallationToken(jwt string, installationID int64) (string, error) {
+	url := fmt.Sprintf("%s/app/installations/%d/access_tokens", githubAPI, installationID)
+
+	req, err := http.NewRequest("POST", url, nil)
```
dagood, line 109: `http.MethodPOST` along the lines of using `http.StatusCreated` below?

gitpr/gitpr.go
```diff
@@ -180,23 +182,38 @@ func (r Remote) GetOwnerSlashRepo() string {
 	return strings.Join(r.GetOwnerRepo(), "/")
 }
 
-// GetUsername queries GitHub for the username associated with a PAT.
-func GetUsername(pat string) string {
-	request, err := http.NewRequest("GET", "https://api.github.com/user", nil)
+// GetUsernameOrAppName returns the username for PAT auth (from /user)
+// or the app name for App-based auth (from /app), based on the useApp flag.
+func GetUsernameOrAppName(auther gitcmd.URLAuther, useApp bool) string {
```
dagood, line 187: Why not just two functions? Or, it seems like this func should be able to figure it out based on what kind of `gitcmd.URLAuther` we get. Or maybe there can be a helper based on the flags that can do the right thing.

go.mod
```diff
@@ -8,6 +8,7 @@ go 1.22.0
 require (
 	github.com/AzureAD/microsoft-authentication-library-for-go v1.3.3
 	github.com/go-test/deep v1.1.1
+	github.com/golang-jwt/jwt/v5 v5.0.0
```
dagood, line 11: Hmm, this is relatively old. Not necessarily a problem, but noticed it.

sync/sync.go
```diff
@@ -72,7 +80,8 @@ func BindFlags(workingDirectory string) *Flags {
 			"The type of Git auth to inject into URLs for fetch/push access. String options:\n"+
 				" none - Leave GitHub URLs as they are. Git may use HTTPS authentication in this case.\n"+
 				" ssh - Change the GitHub URL to SSH format.\n"+
-				" pat - Add the 'github-user' and 'github-pat' values into the URL.\n"),
+				" pat - Add the 'github-user' and 'github-pat' values into the URL.\n"+
+				" app - Add the 'github-app-client-id', 'github-app-installation', and 'github-app-private-key' values into the URL.\n"),
```
dagood, line 84: That isn't really what happens, right?

sync/sync.go
```diff
@@ -268,6 +305,14 @@ func MakeBranchPRs(f *Flags, dir string, entry *ConfigEntry) ([]SyncResult, erro
 		return nil, err
 	}
 
+	var reviewAuther gitcmd.URLAuther
+
+	if f.GitHubPATReviewer != nil {
+		reviewAuther = &gitcmd.GitHubPATAuther{
+			PAT: *f.GitHubPATReviewer,
+		}
+	}
```
dagood, line 314: Possible `nil` use?

dagood (changes requested):

cmd/releasego/get-images-commit.go
gdams: got it, I'll leave the app support in place here then

buildmodel/commands.go
```diff
@@ -139,6 +143,10 @@ func BindPRFlags() *PRFlags {
 		githubPAT:         flag.String("github-pat", "", "Submit the PR with this GitHub PAT, if specified."),
 		githubPATReviewer: flag.String("github-pat-reviewer", "", "Approve the PR and turn on auto-merge with this PAT, if specified. Required, if github-pat specified."),
 
+		gitHubAppClientID:     flag.String("github-app-client-id", "", "Use this GitHub App Client ID to authenticate to GitHub."),
```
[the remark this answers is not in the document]
gdams: The client ID can contain numbers and letters

githubutil/githubutil.go
gdams: I tried this but `http.MethodPost` is a string constant and not a function so not sure it will work?

sync/sync.go
gdams: Copilot failed me :D

sync/sync.go
gdams: Good spot, guessing I can fix this by initialising `reviewAuther` like this:

```go
var reviewAuther gitcmd.URLAuther = gitcmd.NoAuther{}
```

gitpr/gitpr.go
gdams: agreed, it's pretty awkward. I'll split them up

buildmodel/commands.go
```diff
@@ -148,6 +156,18 @@ func BindPRFlags() *PRFlags {
 // submits the resulting commit as a GitHub PR, approves with a second account, and enables the
 // GitHub auto-merge feature.
 func SubmitUpdatePR(f *PRFlags) error {
+	var auther *gitcmd.GitHubPATAuther
+	if f.githubPAT != nil {
+		auther = &gitcmd.GitHubPATAuther{
+			PAT: *f.githubPAT,
+		}
+	}
```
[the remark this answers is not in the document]
gdams: Think I've fixed this now

eng/pipelines/steps/set-bot-git-author.yml
```diff
@@ -5,6 +5,6 @@
 
 steps:
   - script: |
-      git config --global user.name 'microsoft-golang-bot'
-      git config --global user.email 'microsoft-golang-bot@users.noreply.github.com'
+      git config --global user.name 'bot-for-go[bot]'
+      git config --global user.email '199222863+bot-for-go[bot]@users.noreply.github.com'
```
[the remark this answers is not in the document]
gdams: It's the App ID, should I add a comment?

buildmodel/commands.go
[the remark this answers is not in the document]
dagood: I don't think a no-op auther is fundamentally different from letting it try to use nil--it'd fail either way (because auth is required to submit a PR), and arguably it'd be easier to diagnose the problem if it simply attempted to use nil. Maybe the real issue is that this check is too early, and it should be done as part of the "if we have auth" check?

buildmodel/commands.go
```diff
@@ -148,6 +151,18 @@ func BindPRFlags() *PRFlags {
...
+	var reviewAuther gitcmd.URLAuther = gitcmd.NoAuther{}
+	if f.githubPATReviewer != nil {
+		reviewAuther = &gitcmd.GitHubPATAuther{
+			PAT: *f.githubPATReviewer,
+		}
+	}
```
dagood, line 165: Could these be centralized on `GHAuthFlags` to avoid duplication?

buildmodel/commands.go
```diff
@@ -136,8 +134,13 @@ func BindPRFlags() *PRFlags {
 		origin: flag.String("origin", "git@github.com:microsoft/go-images", "Submit PR to this repo. \n[Need fetch Git permission.]"),
 		to:     flag.String("to", "", "Push PR branch to this Git repository. Defaults to the same repo as 'origin' if not set.\n[Need push Git permission.]"),
 
-		githubPAT:         flag.String("github-pat", "", "Submit the PR with this GitHub PAT, if specified."),
-		githubPATReviewer: flag.String("github-pat-reviewer", "", "Approve the PR and turn on auto-merge with this PAT, if specified. Required, if github-pat specified."),
+		GHAuthFlags: GHAuthFlags{
```
dagood, line 137: This should be a function similar to how `BindUpdateFlags` and `sync.BindAzDOVariableFlags` work.

buildmodel/commands.go
```diff
@@ -397,6 +417,14 @@ type UpdateFlags struct {
 	skipSubmoduleUpdate *bool
 }
 
+type GHAuthFlags struct {
+	githubPAT             *string
+	githubPATReviewer     *string
+	gitHubAppClientID     *string
```
dagood, line 423: Consistency?

```suggestion
type GitHubAuthFlags struct {
	gitHubPAT             *string
	gitHubPATReviewer     *string
	gitHubAppClientID     *string
```

buildmodel/commands.go
```diff
@@ -397,6 +417,14 @@ type UpdateFlags struct {
 	skipSubmoduleUpdate *bool
 }
 
+type GHAuthFlags struct {
```
dagood, line 420: Does this really belong in `buildmodel/commands.go`? I'd expect it to live with other GitHub utilities for nice reuse.

dagood (commented):
A few initial comments that might be useful, but seeing now my last comments aren't fully resolved and you weren't the re-review requester, so maybe I should let you cook. 😄

buildmodel/commands.go
dagood: Hmm, but I didn't expect the PAT and PATReviewer to be in one struct. I expected one struct to be used per user/app.

buildmodel/commands.go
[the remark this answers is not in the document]
gdams: I've added an else case to this if statement. I'm not sure it won't be possible for auther to be nil anymore

buildmodel/commands.go
gdams: updated

gitcmd/gitcmd.go
gdams: agreed, updated

go.mod
gdams: agreed, upgraded to v5.2.1

gitcmd/gitcmd.go
```diff
@@ -30,6 +35,40 @@ type URLAuther interface {
...
+			ClientID:       *f.GitHubAppClientID,
+			InstallationID: *f.GitHubAppInstallation,
+			PrivateKey:     *f.GitHubAppPrivateKey,
+		}
+	}
+	return nil
+}
+
+// NewClientFromFlags returns a GitHub client based on the flags (e.g. PAT, GitHub App).
+func NewClientFromFlags(f *githubutil.GitHubAuthFlags, ctx context.Context) (*github.Client, error) {
```
dagood, line 60: This should probably be a method on `githubutil.GitHubAuthFlags`.

gitcmd/gitcmd.go
```diff
@@ -30,6 +35,40 @@ type URLAuther interface {
 	// InsertAuth inserts authentication into the URL and returns it, or if the auther doesn't
 	// apply, returns the url without any modifications.
 	InsertAuth(url string) string
+	// InsertHTTPAuth inserts authentication into the request, if applicable.
+	InsertHTTPAuth(req *http.Request) error
```
dagood, line 39: Missed comment about making a new interface for this? Actually, I think it should even belong in `githubutil`, not `gitcmd`, because this type of auth isn't used for local repositories.

Having the authers centralized in a single package while trying to make them multipurpose might be causing some difficulties. Might want to split the actual implementations out if that helps.

gitcmd/gitcmd.go
```diff
@@ -30,6 +35,40 @@ type URLAuther interface {
...
+		}
+	}
+	if *f.GitHubAppClientID != "" && *f.GitHubAppInstallation != 0 && *f.GitHubAppPrivateKey != "" {
+		return &GitHubAppAuther{
+			ClientID:       *f.GitHubAppClientID,
+			InstallationID: *f.GitHubAppInstallation,
+			PrivateKey:     *f.GitHubAppPrivateKey,
+		}
+	}
+	return nil
```
dagood, line 56: An error (which could be ignored by the caller) would make more sense to me.

gitcmd/gitcmd.go
dagood, line 60: `ctx` should be first by convention.

buildmodel/commands.go
```diff
@@ -117,9 +118,10 @@ type PRFlags struct {
 	origin *string
 	to     *string
 
-	githubPAT         *string
-	githubPATReviewer *string
+	gitHubPAT         *string
+	gitHubPATReviewer *string
 
+	githubutil.GitHubAuthFlags
```
dagood, line 124: Redundant fields?

buildmodel/commands.go
```diff
@@ -214,9 +222,14 @@ func SubmitUpdatePR(f *PRFlags) error {
 	// be strange for this to not be the case and the assumption simplifies the code for now.
 	var existingPR *gitpr.ExistingPR
 
-	if *f.githubPAT != "" {
-		githubUser := gitpr.GetUsername(*f.githubPAT)
-		fmt.Printf("---- User for github-pat is: %v\n", githubUser)
+	if *f.gitHubPAT != "" || *f.GitHubAppClientID != "" {
+		var githubUser string
+		if *f.GitHubAppClientID != "" {
```
dagood, line 227: Could the auther do this, rather than digging into the auther's internals here?

buildmodel/commands.go
```diff
@@ -340,9 +353,13 @@ func SubmitUpdatePR(f *PRFlags) error {
...
 		skipReason = "No origin specified"
-	case *f.githubPAT == "":
-		skipReason = "github-pat not provided"
-	case *f.githubPATReviewer == "":
+	case *f.gitHubPAT == "" && *f.GitHubAppClientID == "":
+		skipReason = "github-pat and github-app-client-id not provided"
+	case *f.GitHubAppClientID != "" && *f.GitHubAppInstallation == 0:
+		skipReason = "github-app-installation not provided"
+	case *f.GitHubAppClientID != "" && *f.GitHubAppPrivateKey == "":
+		skipReason = "github-app-private-key not provided"
```
dagood, line 361: This seems error prone, I'd expect the auther flags abstraction should be able to help.

cmd/releasego/check-limits.go
```diff
@@ -21,14 +22,14 @@ func init() {
...
-	pat := githubutil.BindPATFlag()
+	gitHubAuthFlags := githubutil.BindGitHubAuthFlags()
 
 	if err := p(); err != nil {
 		return err
 	}
 
 	ctx := context.Background()
-	client, err := githubutil.NewClient(ctx, *pat)
+	client, err := gitcmd.NewClientFromFlags(gitHubAuthFlags, ctx)
```
dagood, line 32: I think this diff shows nicely why it makes more sense for this to be a method on `gitHubAuthFlags`, or at least in `githubutil`.

cmd/releasego/get-images-commit.go
```diff
@@ -36,6 +38,7 @@ func handleGetImagesCommit(p subcmd.ParseFunc) error {
 	azdoVarName := flag.String("set-azdo-variable", "", "An AzDO variable name to set to the commit hash using a logging command.")
 	keepTemp := flag.Bool("w", false, "Keep the temporary repository used for polling, rather than cleaning it up.")
 	pollDelaySeconds := flag.Int("poll-delay", 5, "Number of seconds to wait between each poll attempt.")
+	gitHubAuthFlags := *githubutil.BindGitHubAuthFlags()
```
dagood, line 41: Why dereference the flags struct?

cmd/releasego/get-images-commit.go
```diff
@@ -50,6 +53,27 @@ func handleGetImagesCommit(p subcmd.ParseFunc) error {
...
 		return errors.New("no branch specified")
 	}
+
+	if *gitHubAuthFlags.GitHubAppClientID != "" {
+		token, err := githubutil.GenerateInstallationToken(
+			context.Background(),
+			*gitHubAuthFlags.GitHubAppClientID,
+			*gitHubAuthFlags.GitHubAppInstallation,
+			*gitHubAuthFlags.GitHubAppPrivateKey,
+		)
```
dagood, line 63: Shouldn't the auther be doing this?

dagood (changes requested):
Didn't end up having time for a full pass, but some comments on what seem like the core of the changes, at least.

cmd/releasego/report.go
```diff
@@ -105,5 +105,5 @@ func handleReport(p subcmd.ParseFunc) error {
 
 	log.Printf("Reporting %#v\n", s)
 	ctx := context.Background()
-	return buildreport.Update(ctx, owner, name, *pat, *issue, s)
+	return buildreport.Update(ctx, owner, name, *gitHubAuthFlags.GitHubPat, *issue, s)
```
dagood, line 108: This means the command will claim to support app auth, but really it only will accept a PAT.

I would recommend reverting everything other than the usage that is actually being set up in the yaml to keep the PR tidy while working on the factoring.

gitcmd/gitcmd.go
dagood: Add a comment explaining this, and linking the issue?

dagood:
I tried a bit of refactoring--builds, but untested. https://github.com/dagood/go-infra/commit/a6640e1179adb26166496d570bb2ea8e81d4a094

buildmodel/commands.go
```diff
@@ -136,9 +138,8 @@ func BindPRFlags() *PRFlags {
 		origin: flag.String("origin", "git@github.com:microsoft/go-images", "Submit PR to this repo. \n[Need fetch Git permission.]"),
 		to:     flag.String("to", "", "Push PR branch to this Git repository. Defaults to the same repo as 'origin' if not set.\n[Need push Git permission.]"),
 
-		githubPAT:         flag.String("github-pat", "", "Submit the PR with this GitHub PAT, if specified."),
-		githubPATReviewer: flag.String("github-pat-reviewer", "", "Approve the PR and turn on auto-merge with this PAT, if specified. Required, if github-pat specified."),
-
+		gitHubPATReviewer: githubutil.BindPATReviewerFlag(),
+		GitHubAuthFlags:   *githubutil.BindGitHubAuthFlags(),
```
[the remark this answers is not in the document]
gdams: applied your patch

cmd/releasego/report.go
dagood: IMO if WIP makes it hard to review a PR (or update a PR in response to review) or makes any visible change to functionality, it should generally not be committed. I don't see the upside to keeping it in this case.

dagood: I pushed a targeted revert--if any other commands are known not to be complete, I think they should be reverted too.

dagood:
Pushed a few more commits to address comments and fix up a few things, PTAL @gdams

dagood (approved):

PR #218, opened 2025-02-15
Create internal/azurelinux, add local update cmd

dagood:
Create internal/azurelinux package for update logic. To make it work both on GitHub and locally, it reads from a filesystem abstraction that's *almost* `fs.ReadDirFS` and `fs.ReadFileFS`, but not quite. Writes are implemented in the commands themselves.

Being able to update locally is useful for stuff the current automation doesn't support and needs touchups, like major version upgrades.

A loose end: this PR adds some abstraction that I think will help implement major version upgrades in the future and might make updates a little less fragile in certain situations, but major version upgrades aren't implemented at all, and robustness is not tested. Good enough for now--I got the 1.24 upgrade done with it.

Example PR: https://github.com/dagood/azurelinux/pull/22
Identical to old result: https://github.com/microsoft/azurelinux/pull/12307


karianna (approved):

internal/azurelinux/azurelinux.go
```diff
@@ -0,0 +1,451 @@
...
+	v.Spec = specFileRevisionRegex.ReplaceAll(
+		v.Spec,
+		[]byte("${1}"+escapeRegexReplacementValue(newRevisionVersion)),
+	)
+	return nil
+}
+
+func (v *Version) addChangelog(changelogDate time.Time, assets *buildassets.BuildAssets, newVersion, newRelease string) error {
+	const template = `%%changelog
+* %s Microsoft Golang Bot <microsoft-golang-bot@users.noreply.github.com> - %s-%s
```
qmuntal, line 340: If this PR lands after https://github.com/microsoft/go-infra/pull/217, then we should change this line to the following:

```suggestion
* %s bot-for-go[bot] <199222863+bot-for-go[bot]@users.noreply.github.com> - %s-%s
```

qmuntal (approved):
LGTM with some minor comments.

internal/azurelinux/azurelinux.go
dagood: IMO including the username in the code at all is wrong (it should either be a console input or pulled from Git), and that magic number showing up has made it stand out more.

PR #230, opened 2025-03-06
Add go canary as a step to release go pipeline.

mertakman:
This PR enhances the release build pipeline by integrating the 'microsoft-go-canary' build. It introduces steps to initiate the canary build and report its completion status, ensuring a more comprehensive release process. The 'microsoft-go-canary' pipeline ID is hardcoded for directness, avoiding unnecessary indirection that adds obscurity without functional benefit.

### DIFF
--- a/eng/pipelines/steps/release-build-steps.yml
+++ b/eng/pipelines/steps/release-build-steps.yml
@@ -130,6 +130,28 @@ steps:
         start: true
         reason: queued build
 
+    - script: |
+        releasego build-pipeline \
+          -id '1422' \
+          -org 'https://dev.azure.com/dnceng/' \
+          -proj 'internal' \
+          -azdopat '$(System.AccessToken)' \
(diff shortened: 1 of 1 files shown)


eng/pipelines/steps/release-build-steps.yml
```diff
@@ -130,6 +130,28 @@ steps:
         start: true
         reason: queued build
 
+    - script: |
+        releasego build-pipeline \
+          -id '1422' \
```
dagood, line 135: A note for others: this is intentionally hard-coded rather than taken from the variable group (like `-id '$(GoReleaseInnerloopPipelineID)'` above) because the indirection isn't serving any purpose and is more obscure.

dagood (approved):
I'm planning to simply go ahead and include this in the engineering validation revision release we were planning to do for GitHub App auth.

PR #238, opened 2025-03-13
ghapp: migrate dockerupdatepr to use app-based auth

gdams:
Note that I had to update `GetIdentity()` as it was using an installation token when it actually required a JWT token


qmuntal (dismissed):

eng/pipelines/update-images-pipeline.yml
```diff
@@ -73,8 +73,10 @@ extends:
 
               - pwsh: |
                   go run ./cmd/dockerupdatepr `
-                    -origin https://microsoft-golang-bot:$(BotAccount-microsoft-golang-bot-PAT)@github.com/microsoft/go-images `
```
gdams, line 76: This pipeline is disabled and no longer run, I'm actually tempted to delete this yml pipeline altogether and delete the pipeline in AzDo? WDYT @dagood?

dagood: Yeah, I doubt we'll be using it again any time soon.

githubutil/auth.go
```diff
@@ -129,10 +129,15 @@ func (a GitHubAppAuther) InsertHTTPAuth(req *http.Request) error {
 
 func (a GitHubAppAuther) GetIdentity() (string, error) {
 	ctx := context.Background()
-	client, err := NewInstallationClient(ctx, a.ClientID, a.InstallationID, a.PrivateKey)
```
dagood, line 132: Make a new func called `NewAppClient` that explains the difference in its comment? Then `GenerateInstallationToken` can share it.

dagood, line 132: It doesn't look like `NewInstallationClient` is actually used outside the package, and don't need to be exported. Same with hypothetical `NewAppClient`, and `NewClient` can also go away once all usage is migrated.

buildmodel/commands.go
```diff
@@ -190,6 +190,17 @@ func SubmitUpdatePR(f *PRFlags) error {
 		Purpose: "auto-update",
 	}
 
+	// if using github app we need to insert the app token into the URL
+	// so that the app can push to the repo
+	if *f.Auth.GitHubAppClientID != "" {
```
dagood, line 195: Why the `if`? Isn't this the case for PATs too, now that it isn't being injected into `origin`?

dagood (changes requested):

buildmodel/commands.go
gdams: agreed, removed the if

githubutil/auth.go
gdams: done

eng/pipelines/update-images-pipeline.yml
gdams: I've deleted the file

githubutil/auth.go
gdams: I've changed `NewInstallationClient` to `newInstallationClient`

buildmodel/commands.go
```diff
@@ -221,6 +221,8 @@ func SubmitUpdatePR(f *PRFlags) error {
 		}
 		fmt.Printf("---- Skipping finding PR to update; no auth: %v\n", err)
 	} else {
+		*f.origin = auther.InsertAuth(*f.origin)
+		*f.to = auther.InsertAuth(*f.to)
```
dagood, line 225: I looked at this again in the broader context, and this is actually a good opportunity to stop passing around the auth as unclearly. `auther.InsertAuth` should be used as late as possible (generally speaking) to make it clearer to future readers where sensitive information might be. This helps avoid inadvertently leaking it in a future change (e.g. in the PR desc).

buildmodel/commands.go
```diff
@@ -328,7 +328,7 @@ func SubmitUpdatePR(f *PRFlags) error {
 	runOrPanic(newGitCmd("commit", "-m", commitMessage))
 
 	// Push the commit.
-	args := []string{"push", *f.origin, b.PRBranchRefspec()}
+	args := []string{"push", auther.InsertAuth(*f.origin), b.PRBranchRefspec()}
```
dagood, line 331: Missing auth for `runOrPanic(newGitCmd("fetch", "--no-tags", *f.to, b.PRBranchRefspec()))` and `runOrPanic(newGitCmd("fetch", "--no-tags", *f.origin, b.BaseBranchFetchRefspec()))` calls earlier?

dagood, line 331: What if this is a dry run without auth?

Maybe assign `auther` to be a `NoAuther` in the error case to avoid nils?

githubutil/auth.go
```diff
@@ -148,16 +150,25 @@ func (a GitHubAppAuther) getInstallationToken() (string, time.Time, error) {
 		a.PrivateKey)
 }
 
-func GenerateInstallationToken(ctx context.Context, clientID string, installationID int64, privateKey string) (string, time.Time, error) {
+// newAppClient creates a new JWT-based GitHub client using the provided client ID and private key.
+func newAppClient(clientID, privateKey string) (*github.Client, error) {
```
dagood, line 154: Put this func next to `newInstallationClient` and accept `ctx`?

githubutil/auth.go
```diff
@@ -148,16 +150,25 @@ func (a GitHubAppAuther) getInstallationToken() (string, time.Time, error) {
...
 
 	tokenSource := oauth2.StaticTokenSource(&oauth2.Token{AccessToken: jwt})
-	tokenClient := oauth2.NewClient(ctx, tokenSource)
+	tokenClient := oauth2.NewClient(context.Background(), tokenSource)
 
 	client := github.NewClient(tokenClient)
+	return client, nil
+}
+
+func GenerateInstallationToken(ctx context.Context, clientID string, installationID int64, privateKey string) (string, time.Time, error) {
```
dagood, line 167: I don't think this should be exported either.

dagood (commented):
Took another look at the `githubutil` changes.

sync/sync.go
```diff
@@ -686,15 +686,15 @@ func MakeBranchPRs(f *Flags, dir string, entry *ConfigEntry) ([]SyncResult, erro
...
 
-		auther, err := f.Auth.NewAuther()
+		auther, err := f.Auth.NewAuther(*f.DryRun)
 		if err != nil {
 			if !errors.Is(err, githubutil.ErrNoAuthProvided) {
 				return nil, err
 			}
 			c.SkipReason = fmt.Sprintf("provided auth doesn't support PR submission: %v", err)
 			continue
 		}
```
dagood, line 696: The bool changes behavior pretty significantly--I think this block is the clearest at showing why a `githubutil.ErrNoAuthProvided` must happen even in a dry run. The command itself needs to determine what to do about a dry run: sometimes a lot of code needs to be skipped that can't be done anonymously at all (whether due to rate limits or permissions).

This is more of what I was thinking: https://github.com/dagood/go-infra/commit/898934329d5b2981225905071e22a58d489c50db

gdams: agreed, I've applied your commit, let me know what you think

gdams:
@dagood I had to push a handful of unrelated linter fixes to this PR to get the linter run green. These are all pretty sane changes IMO so it feels safe enough to do this as part of this change

dagood:
IMO that should be its own PR--we aren't in a rush. The totally uncontroversial way to fix it would be to pin golangci-lint. It shouldn't have been set up in a way that it can float onto a new major version. (But that fix should also be its own PR.)

gdams:
> IMO that should be its own PR--we aren't in a rush. The totally uncontroversial way to fix it would be to pin golangci-lint. It shouldn't have been set up in a way that it can float onto a new major version. (But that fix should also be its own PR.)

Good point, I've raised https://github.com/microsoft/go-infra/pull/241, once merged I'll rebase this PR

gdams:
@dagood I've rebased this now to remove the linter fixes

githubutil/auth.go
dagood: (This popped up again.)

githubutil/auth.go
dagood: (This popped up again.)

dagood (commented):
Looks like a couple changes got lost.

gdams:
> Looks like a couple changes got lost.

good spot, fixed it up now

dagood (approved):
Thanks!

PR #255, opened 2025-05-05
Implement Application Insights package

qmuntal:
This PR contains a trimmed down version of https://github.com/microsoft/ApplicationInsights-Go.
It is tailored for the use of the [Microsoft build of Go](https://github.com/microsoft/go).

These are the changes made to the original package:

- Remove all external dependencies.
- Remove all telemetry types except for `Event`.
- Simplify the API.
- Modernize the code.
- Telemetry events no longer have a unique operation UUID (we don't care about it).
- A bug has been fixed that leaked a goroutine when closing the client.
- The API is now more similar to the one in https://github.com/golang/telemetry.
- The time mocking for tests now uses the experimental `tetsing/synctest` package instead of an external library (requires Go 1.24). This simplified the code and makes testing more robust (it found the aforementioned bug). See https://go.dev/blog/synctest for more info.
- The code is more performant now
- Test have been expanded and improved.

A side benefit is that now vendoring this module only takes 1k LoC instead of the previous 17k LoC.

Usage example:

```go
package main

import (
  "context"

  "github.com/microsoft/go-infra/telemetry/internal/appinsights"
)

var client = &appinsights.Client{InstrumentationKey: "fake_key"}

var counter = client.NewEvent("test event")

func main() {
  counter.Inc()
  client.TrackEvent("another event")
  client.Close(context.Background())
}
```

This are the `TrackEvent` call performance improvements taking the official SDK as baseline:

```
goos: windows
goarch: amd64
pkg: github.com/microsoft/go-infra/telemetry/internal/appinsights
cpu: Intel(R) Core(TM) i7-10850H CPU @ 2.70GHz
                          │   old.txt    │              new.txt               │
                          │    sec/op    │   sec/op     vs base               │
ClientBurstPerformance-12   4028.0n ± 9%   867.7n ± 9%  -78.46% (p=0.001 n=7)

                          │   old.txt   │              new.txt              │
                          │    B/op     │    B/op     vs base               │
ClientBurstPerformance-12   3004.0 ± 0%   748.0 ± 3%  -75.10% (p=0.001 n=7)

                          │   old.txt   │              new.txt              │
                          │  allocs/op  │ allocs/op   vs base               │
ClientBurstPerformance-12   21.000 ± 0%   2.000 ± 0%  -90.48% (p=0.001 n=7)
```


qmuntal:
I've bumped the Ubuntu image to 22.04, as 20.04 is going to be removed soon and it is failing now (see https://github.com/actions/runner-images/issues/11101).

telemetry/internal/appinsights/client.go
```diff
@@ -0,0 +1,184 @@
...
+	e.Add(1)
+}
+
+// Add adds n to the counter. n cannot be negative, as counts cannot decrease.
+func (e *Event) Add(n uint64) {
+	if n == 0 {
+		return
+	}
+	for range n {
+		e.client.track(contracts.EventData{
```
dagood, line 179: It feels like there should be some practical limit here because of the wasteful-seeming way they're queued up to be transmitted... maybe at least some caution in the doc comment.

telemetry/internal/appinsights/client_test.go
```diff
@@ -0,0 +1,102 @@
...
+		if err != nil {
+			t.Fatalf("Coudln't create gzip reader: %s", err.Error())
+		}
+
+		// Read payload
+		body, err := io.ReadAll(reader)
+		reader.Close()
+		if err != nil {
+			t.Fatalf("Couldn't read compressed data: %s", err.Error())
+		}
```
dagood, line 86: Seems like a decent chunk of this could be shared with `telemetry/internal/appinsights/transmitter_test.go`, up to the GZIP magic number?

telemetry/internal/appinsights/client.go
```diff
@@ -0,0 +1,184 @@
+package appinsights
```
dagood, line 1: Copyright headers?

telemetry/internal/appinsights/inmemorychannel.go
```diff
@@ -0,0 +1,453 @@
...
+	"net/http"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/microsoft/go-infra/telemetry/internal/appinsights/internal/contracts"
+)
+
+var (
+	submit_retries = []time.Duration{time.Duration(10 * time.Second), time.Duration(30 * time.Second), time.Duration(60 * time.Second)}
```
dagood, line 14:
```suggestion
	submitRetries = []time.Duration{time.Duration(10 * time.Second), time.Duration(30 * time.Second), time.Duration(60 * time.Second)}
```

telemetry/internal/appinsights/inmemorychannel.go
```diff
@@ -0,0 +1,453 @@
...
+	retry bool
+
+	// If retrying, what is the max time to wait before finishing up?
+	timeout time.Duration
+
+	// If specified, a message will be sent on this channel when all pending telemetry items have been submitted
+	callback chan struct{}
+}
+
+// newInMemoryChannel creates an inMemoryChannel instance and starts a background submission goroutine.
```
dagood, line 50: Maybe mention what's necessary to make sure you don't leak a goroutine after getting a result from this func?

telemetry/internal/appinsights/inmemorychannel.go
```diff
@@ -0,0 +1,453 @@
...
+		state.channel.waitgroup.Add(1)
+
+		// If we have a callback, wait on the waitgroup now that it's
+		// incremented.
+		state.channel.signalWhenDone(state.callback)
+
+		go func(buffer []*contracts.Envelope, retry bool, retryTimeout time.Duration) {
+			defer state.channel.waitgroup.Done()
+			state.channel.transmitRetry(buffer, retry, retryTimeout)
+		}(state.buffer, state.retry, state.retryTimeout)
```
dagood, line 296: Personally I find capture more concise/readable/maintainable, although I do see the explicit func args pattern often in Go code. Curious if there is a deeper reason for it that I'm missing vs.:

```suggestion
		buffer := state.buffer
		retry := state.retry
		retryTimeout := state.retryTimeout
		go func() {
			defer state.channel.waitgroup.Done()
			state.channel.transmitRetry(buffer, retry, retryTimeout)
		}()
```

(I can't say for sure I follow why these are all copied, but I have an intuition about why it *could* be important...)

telemetry/internal/appinsights/inmemorychannel.go
```diff
@@ -0,0 +1,453 @@
...
+			return
+		}
+
+		// Check for throttling
+		if result.isThrottled() {
+			if !result.retryAfter.IsZero() {
+				channel.throttle.retryAfter(result.retryAfter)
+			} else {
+				// TODO: Pick a time
+			}
```
dagood, line 407: TODO spotted. I guess currently this means that a timeless throttle doesn't delay the retry at all?

telemetry/internal/appinsights/package.go
```diff
@@ -0,0 +1,8 @@
+// Package appinsights provides an interface to submit telemetry to Application Insights.
+// See more at https://azure.microsoft.com/en-us/services/application-insights/
```
dagood, line 2: This redirects to the Azure Monitor product page, perhaps a more useful link:

```suggestion
// Package appinsights provides an interface to submit telemetry to Application Insights,
// a component of Azure Monitor. This package calls the Classic API.
// See https://learn.microsoft.com/en-us/azure/azure-monitor/app/api-custom-events-metrics
```

telemetry/internal/appinsights/throttle.go
```diff
@@ -0,0 +1,142 @@
...
+type throttleMessage struct {
+	query     bool
+	wait      bool
+	throttle  bool
+	stop      bool
+	timestamp time.Time
+	result    chan bool
+}
+
+func newThrottleManager() *throttleManager {
```
dagood, line 20: Would be nice to have some documentation about the intent of this struct and its methods.

dagood (commented):
Nice. I didn't poke at the test code much.

telemetry/internal/appinsights/inmemorychannel.go
qmuntal: Fixed. The retry will use the `Client.MaxBatchInterval` as delay.

telemetry/internal/appinsights/inmemorychannel.go
qmuntal: Yeah, I also dislike this code. Will improve it 😄.

telemetry/internal/appinsights/client.go
qmuntal: Good catch. Added.

telemetry/internal/appinsights/client.go
qmuntal: I've optimized this method a bit. I still have to add tests for this method to see if how well it behave.

telemetry/internal/appinsights/inmemorychannel.go
qmuntal: Good point. I've moved the `acceptLoop` call outside of this constructor to let the caller control when the loop is started.

telemetry/internal/appinsights/throttle.go
qmuntal: I've removed this struct altogether. Too complicated with little benefit.

qmuntal:
This PR is ready for review!

gdams (approved):

telemetry/internal/appinsights/inmemorychannel.go
```diff
@@ -0,0 +1,311 @@
...
+	}
+
+	channel.closed.Store(true)
+	channel.cancelCauseFunc(errStopped)
+}
+
+// Flushes and tears down the submission goroutine and closes internal
+// channels.  Returns a channel that is closed when all pending telemetry
+// items have been submitted and it is safe to shut down without losing
+// telemetry.
```
dagood, line 119: Doesn't seem accurate?

```suggestion
// close flushes and tears down the submission goroutine and closes internal
// channels. Returns when all pending telemetry items have been submitted
// (it is then safe to shut down without losing telemetry) or when
// the context is canceled.
```

telemetry/internal/appinsights/inmemorychannel.go
```diff
@@ -0,0 +1,311 @@
...
+			channel.cancelCauseFunc(errClosed)
+		}
+		return
+	}
+
+	// Copy the items to a temporary buffer to let the caller
+	// reuse the item slice.
+	buf := itemsBuf.Get().(*[]batchItem)
+	*buf = (*buf)[:0]
+	*buf = append(*buf, items...)
```
dagood, line 244: I'm vaguely aware of [a concern with `sync.Pool` usage for buffers](https://github.com/golang/go/issues/23199) where if you sometimes grow the buffers, you might end up accumulating lots of buffers that are all needlessly high-capacity, putting pressure on memory. Currently each of these bufs have 1024 cap, and if `items` is longer than that, it looks like this buf would grow, running into that gotcha.

I don't know for sure if that concern applies here or not, but if so, perhaps we should e.g. grab multiple bufs from the pool if `len(items)` > 1024 rather than growing the buf we happen to get?

telemetry/internal/appinsights/internal/contracts/envelope.go
```diff
@@ -0,0 +1,105 @@
...
+	// https://en.wikipedia.org/wiki/ISO_8601#UTC. Note: the number of decimal
+	// seconds digits provided are variable (and unspecified). Consumers should
+	// handle this, i.e. managed code consumers should not use format 'O' for
+	// parsing as it specifies a fixed length. Example:
+	// 2009-06-15T13:45:30.0000000Z.
+	Time time.Time `json:"time"`
+
+	// Sampling rate used in application. This telemetry item represents 1 /
+	// sampleRate actual telemetry items.
+	SampleRate float64 `json:"sampleRate"`
```
dagood, line 52: This is confusing... this telemetry item isn't an "actual" telemetry item?

(Also, more non-go-idiom starts of comments throughout the file.)

telemetry/internal/appinsights/transmitter.go
```diff
@@ -0,0 +1,167 @@
...
+
+	// Compress the payload
+	var postBody bytes.Buffer
+	gzipWriter := gzip.NewWriter(&postBody)
+	if _, err := gzipWriter.Write(payload); err != nil {
+		gzipWriter.Close()
+		return nil, fmt.Errorf("failed to compress the payload: %v", err)
+	}
+
+	gzipWriter.Close()
```
dagood, line 71: Not handling `gzipWriter.Close()` error?

telemetry/internal/appinsights/inmemorychannel.go
qmuntal: Good point. The `items` size is capped to `Client.MaxBatchSize` (defaults to 1024), so the polled buffer sizes can't grow unbounded and cause an OOM, which is the main complain in that issue. I've added a comment to note this limit, and also moved the pool from being a global variable to being a property of the channel so that we can allocate buffers knowing for sure the maximum capacity.

telemetry/internal/appinsights/internal/contracts/envelope.go
qmuntal: I've fixed the non-idiomatic comments. About the comment, this comes directly from the API documentation, and is the same in other sdks, so I wouldn't touch it. The phrase is confusing, but I think it refers to the fact that a single envelope can be the result of sampling multiple items. In fact I might be able to use this to implement `Event.Add` where `n` is higher than 1. I'll give it a try once I can upload telemetry to a dev appinishgts endpoint.

dagood (approved):

PR #261, opened 2025-05-26
Implement telemetry API

qmuntal:
This PR exposes the necessary APIs to upload telemetry from the Microsoft build of Go. The API is modeled in a way that can easily be used together with https://github.com/golang/telemetry so that both can coexist.

Usage example:

```go
import (
  "context"

  "github.com/microsoft/go-infra/telemetry"
  "github.com/microsoft/go-infra/telemetry/counter"
)

func main(
  telemetry.Start(telemetry.Config{
    InstrumentationKey: "fake_key",
  })
  counter.Inc("fake_counter")
  telemetry.Close(context.Background())
)
```

Only counters that exist in the `telemetry/config.json` file will be uploaded (the file will be embedded in any binary importing `telemetry`). Each counter comes with 5 metadata entries: GOOS, GOARCH, Go version, program name, and a unique session id.

The philosophy here is the same as in upstream Go: we don't upload data that does not appear verbatim in the config file. This means that the config file should contain all supported GOOs, GOARCHs, Go versions, and program names. It will simplify the CELA approval request.

For now, the list only includes one counters: the number of `cmd/go` invocations (which is the Go toolchain entry point). We can extend this list as we find other counters that are valuable for us. Note that this will make the telemetry upload work almost unnoticeable given that it will only upload one counter at the beginning of the program execution.

For https://github.com/microsoft/go/issues/1672.

### DIFF
--- a/eng/pipelines/pr-pipeline.yml
+++ b/eng/pipelines/pr-pipeline.yml
@@ -20,11 +20,11 @@ jobs:
       - template: steps/set-bot-git-author.yml
 
       - script: |
-          go test -json ./... > $(Build.StagingDirectory)/TestResults.json
+          go test -json ./... >> $(Build.StagingDirectory)/TestResults.json
         displayName: Test
         
       - script: |
-          go test -json -race ./... > $(Build.StagingDirectory)/TestResults.json
+          go test -json -race ./... >> $(Build.StagingDirectory)/TestResults.json
         displayName: Test Telemetry
         workingDirectory: $(Build.SourcesDirectory)/telemetry
         env:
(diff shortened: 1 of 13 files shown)


gdams (approved):

dagood:
I missed that these write to the same file (not append)--probably not related to the CI failing, but might be a good time to fix:

https://github.com/microsoft/go-infra/blob/d881189be0ad24f773adfbcdb0ecd4b4aab9352a/eng/pipelines/pr-pipeline.yml#L22-L31

gdams (approved):

PR #265, opened 2025-05-29
Add arbitrary telemetry client package, properties

dagood:
Add a package for arbitrary events that can be imported separately, avoiding importing the config system and counter APIs, and avoiding exposing the arbitrary event APIs to the Microsoft build of Go.

Add support for `Properties` to the events. This is pretty simple: just an extra map pointer that's always nil for the Microsoft build of Go cases.

Includes properties map sanitization, copied from the original module. Made a simple test to check that it's working.

Doesn't include features like CommonProperties attached to the Client--this can be done just as well (and perhaps better) by the caller. No metrics, only events.

No effect on the benchmark:

```
goos: windows
goarch: amd64
pkg: github.com/microsoft/go-infra/telemetry/internal/appinsights
cpu: Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz
                         │ before.txt  │           after.txt           │
                         │   sec/op    │   sec/op     vs base          │
ClientBurstPerformance-8   695.6n ± 2%   693.4n ± 2%  ~ (p=0.853 n=10)
```

### DIFF
--- a/telemetry/internal/appinsights/client.go
+++ b/telemetry/internal/appinsights/client.go
@@ -108,6 +108,14 @@ func (c *Client) TrackEvent(name string) {
 	c.NewEvent(name).Inc()
 }
 
+func (c *Client) TrackEventWithProperties(name string, properties map[string]string) {
+	c.track(contracts.EventData{
+		Name:       name,
+		Ver:        2,
+		Properties: properties,
+	}, 1)
+}
+
 // Forces the current queue to be sent.
 func (c *Client) Flush() {
 	if !c.initialized.Load() {
--- a/telemetry/README.md
+++ b/telemetry/README.md
@@ -1,3 +1,7 @@
 # telemetry
 
-This directory contains the telemetry transmission code for the Microsoft build of Go.
+This directory, the `telemetry` package, contains the telemetry transmission code for the Microsoft build of Go.
+It is specialized to work similarly to the upstream telemetry counters.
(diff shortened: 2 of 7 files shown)


telemetry/internal/appinsights/client.go
```diff
@@ -108,6 +108,14 @@ func (c *Client) TrackEvent(name string) {
 	c.NewEvent(name).Inc()
 }
 
+func (c *Client) TrackEventWithProperties(name string, properties map[string]string) {
```
qmuntal, line 111: There is no need to add a new method, just add the `properties` parameter the `TrackEvent`. That shouldn't have any performance impact.

Also, in the Go toolchain I'm using a nil Client to denote that uploading is disabled. Note that `Event.Add` does nothing if `Event.client` is nil. This case is not handled here. You can easily fix it by moving the client nil check short-circuit from `Event.Add` to `Client.track`.

qmuntal (changes requested):
LGTM minus once comment.

telemetry/internal/appinsights/client.go
dagood: Will do, this was just to avoid touching the existing API for now--there are a decent number more places that need to be touched to pass this through the `Event` type.

Ah, didn't notice the nil receivers.

qmuntal (approved):

PR #291, opened 2025-07-23
telemetry: replace Client.ErrorLog for a structured logger and default to not log anything

qmuntal:
The current way of handling telemetry errors is tricky to use by callers:

- By default it logs into `os.StdErr`, which is unexpected in most cases.
- The log prefixes and flags are hardcoded and are difficult to compose with caller-provided loggers.
- Some telemetry errors are not really errors, but warnings, and callers can't differentiate between them.
- There is no way to log upload statistics.

This PR does the following breaking changes to improve the telemetry logging story:

- `Client.ErrorLog` has been removed.
- `Client.Logger` has been added. It is of type [slog. Logger](https://pkg.go.dev/log/slog#Logger). This allows logging messages with different levels of severity. It opens the door for also logging upload statistics.
- If the logger is nil, nothing is logged.

While here, some missing error cases have been added, some error messages have been improved, and more error-related tests have been added.

Fixes https://github.com/microsoft/go-lab/issues/219. See https://github.com/microsoft/go-infra/pull/291/commits/a954cede40cdf14973f2dffd5711f6eccf3c953d.

### DIFF
--- a/telemetry/internal/appinsights/inmemorychannel.go
+++ b/telemetry/internal/appinsights/inmemorychannel.go
@@ -314,7 +314,7 @@ func (channel *inMemoryChannel) transmitRetry() bool {
 	}
 
 	// If the response is not successful, check if we can retry.
-	succeed, failed, retryItems := resp.result(*itemsPtr)
+	succCount, failCount, retryItems := resp.result(*itemsPtr)
 
 	// Remove items that have been retried too many times.
 	retries := len(retryItems)
@@ -326,9 +326,9 @@ func (channel *inMemoryChannel) transmitRetry() bool {
 	}
 
 	dropped := retries - len(retryItems)
-	channel.inflight.Add(-int64(succeed + failed + dropped))
-	if failed > 0 {
-		channel.error("server rejected items", "itemsLost", failed, "statusCode", resp.statusCode)
+	channel.inflight.Add(-int64(succCount + failCount + dropped))
+	if failCount > 0 {
+		channel.error("server rejected items", "itemsLost", failCount, "statusCode", resp.statusCode)
 	}
 	if dropped > 0 {
 		channel.error("items dropped due to retry limit", "itemsLost", dropped)


telemetry/internal/appinsights/inmemorychannel.go
```diff
@@ -80,8 +82,16 @@ func newInMemoryChannel(endpointUrl string, batchSize int, batchInterval time.Du
 	return channel
 }
 
-func (channel *inMemoryChannel) logf(format string, args ...any) {
-	channel.errorLog.Printf(format, args...)
+func (channel *inMemoryChannel) error(err error) {
+	channel.errorsMu.Lock()
+	defer channel.errorsMu.Unlock()
+	channel.errors = append(channel.errors, err)
```
dagood, line 88: Does this have any limit? Otherwise it can be a pseudo-memory-leak for a long-running application (like my use case).

qmuntal: I logged in to mark this PR as not ready for review just for this reason 😓. Thanks!

qmuntal: Ok, I repurposed this PR with a completely different approach (see new PR description).

telemetry/internal/appinsights/client.go
```diff
@@ -72,15 +71,10 @@ func (c *Client) init() {
...
-		if errorLog == nil {
-			std := log.Default()
-			errorLog = log.New(std.Writer(), std.Prefix()+"appinsights: ", std.Flags())
-		}
-		c.channel = newInMemoryChannel(endpoint, batchSize, batchInterval, httpClient, errorLog)
+		c.channel = newInMemoryChannel(endpoint, batchSize, batchInterval, httpClient, c.Logger)
 		c.context = setupContext(c.InstrumentationKey, c.Tags)
 		if err := contracts.SanitizeTags(c.context.Tags); err != nil {
-			c.channel.logf("Warning sanitizing tags: %v", err)
+			c.channel.warn("tags have been sanitized", "error", err)
```
dagood, line 77: Even if no changes were made, I would still say that tags "have been sanitized" by passing through the sanitizer. This is more like "tags were not sanitary and have been sanitized" or something like that.

telemetry/internal/appinsights/inmemorychannel.go
```diff
@@ -269,42 +297,30 @@ func (channel *inMemoryChannel) transmitRetry() bool {
...
 	}
 
 	// Filter down to failed items.
-	retryItems := result.getRetryItems(*itemsPtr)
-
-	// Delete items that have been retried more than 2 times.
-	retryItems = slices.DeleteFunc(retryItems, func(item batchItem) bool {
-		return item.retries > 2
-	})
+	succed, failed, retryItems := result.getRetryItems(*itemsPtr)
```
dagood, line 316:
```suggestion
	succeed, failed, retryItems := result.getRetryItems(*itemsPtr)
```

telemetry/internal/appinsights/transmitter.go
```diff
@@ -142,28 +138,45 @@ func canRetryBackendError(berror contracts.BackendResponseError) bool {
 		berror.StatusCode == tooManyRequestsOverExtendedTimeResponse
 }
 
-func (result *transmissionResponse) getRetryItems(items []batchItem) []batchItem {
+func (result *transmissionResponse) getRetryItems(items []batchItem) (succed, failed int, retries []batchItem) {
```
dagood, line 141:
```suggestion
func (result *transmissionResponse) getRetryItems(items []batchItem) (succeed, failed int, retries []batchItem) {
```

telemetry/internal/appinsights/transmitter.go
```diff
@@ -142,28 +138,45 @@ func canRetryBackendError(berror contracts.BackendResponseError) bool {
...
+	defer func() {
+		for i := len(retries) - 1; i >= 0; i-- {
+			item := &retries[i]
+			// If the item has been retried too many times, we consider it failed.
+			if item.retries > 2 {
+				retries = retries[:i]
```
dagood, line 147: This seems complicated... I'm not totally sure what `getRetryItems` is intending to do overall, and what assumptions it has. E.g. is this intended to remove item `i` and all items after it, even if some of the items after `i` had retries remaining? Or is this not a possible situation to get into?

I feel like doc comments and/or more expressive helpers are needed here.

telemetry/internal/appinsights/transmitter.go
```diff
@@ -138,30 +138,20 @@ func canRetryBackendError(berror contracts.BackendResponseError) bool {
...
-				failed++
-				continue
-			}
-			item.retries++
-		}
-	}()
-	if result.statusCode == partialSuccessResponse {
+// result returns the number of succeeded and failed items, and a list of items that can be retried.
+// Items is the complete list of result that was sent.
+func (resp *transmissionResponse) result(items []batchItem) (succeed, failed int, retries []batchItem) {
```
dagood, line 143: Ah, I think I get what you're going for with `succed` now. It normally reads to me as simply a misspelling or as "sucked" with unusual spelling (maybe just a US slang thing?), but I think I've seen this as `succCount` (and `failCount`) before elsewhere. I think that would be readable enough with the context of `Count`.

(Could also go with `succeeded` (like the comment) to fit tense with `failed`, but I get abbreviating.)

I think `retries` would also be clearer as `retryItems` like it's actually named at the call site--these aren't retries yet, just items that can/should be retried. Perhaps `rs` is a fine abbreviation that naturally directs more attention to the doc comment where it's spelled out (and the call site can keep its longer name for the local var).

dagood (approved):

telemetry/internal/appinsights/transmitter.go
qmuntal: Thanks, make sense!

PR #316, opened 2025-08-12
gomoddependabot: add tool to manage dependency groups

dagood:
Add simple tool to update `.github/dependabot.yml` to point at each go.mod file in the repository. Two purposes:

* Make sure every go.mod actually receives updates.
* Group such that each go.mod gets its own update PR, but multiple dependency updates can occur within each PR, reducing human PR load.

Use JSON in the yml file to avoid actually dealing with a full-fledged yml parser and yml round-trip for the unrelated lines.

Haven't confirmed how the generated `.github/dependabot.yml` behaves.

Force `\n`. This might require a `.gitattributes` change for 100% behavior repro on Windows, but this is probably not important. (It's common for Go tools to force `\n`, anyway, so devs on Windows should recognize the situation when it shows up in Git.)


karianna (approved):

gdams (approved):

dagood:
I think a good way to test this would be to: next time we get a bunch of dependabot PRs for Go dependencies in one of our repos, instead of merging them, apply this tool and merge that first. That should trigger dependabot to re-evaluate, and we'll see what it ends up actually doing when we get a lot of dependency updates at once.

PR #338, opened 2025-09-16
Modernize pass

dagood:
Run `go run golang.org/x/tools/gopls/internal/analysis/modernize/cmd/modernize@latest -fix -test ./...`

Nothing huge, but was curious what it would show for this repo after seeing a handful of upstream issues about this tool, and might as well submit it. Using `strings.SplitSeq` as a modernization is an interesting one that wasn't on my radar.


qmuntal (approved):
Nice

gdams (approved):

PR #458, opened 2026-04-03
Fix inlinerange to produce mapping output in mapping context

dagood:
Quick Copilot fix to support some data reuse in our pipelines.

Specifically, lets us unify:

https://github.com/microsoft/go/blob/8b6d5c3ca87e2f493caad3e4eaa86f2ac0384311/eng/pipeline/pr-pipeline.yml#L23-L30

https://github.com/microsoft/go/blob/8b6d5c3ca87e2f493caad3e4eaa86f2ac0384311/eng/pipeline/rolling-innerloop-pipeline.yml#L46-L53

---

`inlinerange` always produces a sequence (list), even when used as a mapping value where map entries are expected:

```yaml
containers:
  ${ inlinerange .containers }:
    ${ .container }:
      image: ${ .image }
```

Produces `- ubuntu2204:` (sequence items) instead of `ubuntu2204:` (map entries).

### Root cause

`evalRangeResult` unconditionally returns a `SequenceNode`. When the parent mapping dissolves into this SequenceNode, map-valued contexts get sequence output.

### Fix

- Added `inSequence` flag on `evalRange`, set when the range's parent mapping is a sequence item (detected during `eval()` of `SequenceNode`)
- Added `evalRangeFlatten`: when `!inSequence` and all body iterations produce `MappingNode`s, flattens them into a single combined `MappingNode`
- Existing sequence-context behavior (`- ${ inlinerange }:`) is unchanged — `inSequence=true` preserves the `SequenceNode` path

### Test

Added `inlinerange-map-body` test case exercising the mapping-value context.

### DIFF
--- a/internal/pipelineymlgen/eval.go
+++ b/internal/pipelineymlgen/eval.go
@@ -800,8 +800,9 @@ type evalRange struct {
 	collection any
 	body       *yaml.Node
 	// inSequence is true when this range is inside a mapping that is a
-	// sequence item. In that case each iteration produces a separate sequence
-	// element. When false (mapping value context), iterations that produce
-	// MappingNodes are flattened into a single combined MappingNode.
+	// sequence item. When true, the range always produces a SequenceNode
+	// (preserving list item structure). When false (mapping value context),
+	// iterations that produce MappingNodes are flattened into a single
+	// combined MappingNode.
 	inSequence bool
 }
--- a/internal/pipelineymlgen/eval.go
+++ b/internal/pipelineymlgen/eval.go
@@ -225,6 +225,15 @@ func (e *EvalState) eval(orig *yaml.Node) (any, error) {
 			if err != nil {
 				return fail(fmt.Errorf("evaluating sequence item: %w", err))
 			}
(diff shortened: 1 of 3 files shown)


dagood:
I'm realizing that `inlinerange` is underspecified... e.g. I don't think there's a right answer (or very intuitive answer) as to which output this:

```yml
- ${ inlinerange (list "a" "b") }:
  - ${ yml . }: ex
```

Should result in:

```yml
# 1
- a: ex
- b: ex
```

```yml
# 2
- - a: ex
  - b: ex
```

With a map inside, similar oddity:

```yml
- ${ inlinerange (list "a" "b") }:
    ${ yml . }: ex
```

```yml
# 3
- a: ex
- b: ex
```

```yml
# 4
- a: ex
  b: ex
```

First impression is that because `inlinerange` says "inline" it should try to inline (merge) each output element with the parent element. So: 1 and 3.

Perhaps a non-inlining version could be `ymlrange`, since we have a `yml` expression already. 2 and 4.

---

On second thought... `inlinerange` -> 1 and 4 is intuitive. The others can be expressed with some extra syntax that's also intuitive.

PR #462, opened 2026-04-06
pipelineymlgen: make inlinerange more consistent and usable

dagood:
Make the input requirements of `pipelineymlgen` more consistent and make the output more predictable. The current implementation tends to over-inline or result in a sequence when a map is needed, making some output impossible to achieve.

* Improves on https://github.com/microsoft/go-infra/pull/443


internal/pipelineymlgen/eval.go
```diff
@@ -460,9 +469,16 @@ func (e *EvalState) mappingToYAML(m *evalMapping) (*yaml.Node, error) {
...
+				} else {
+					return fail(fmt.Errorf("mapping contains multiple kinds: had %v, now also %v", kindStr(n), kindStr(node)))
+				}
+				n.Content = append(n.Content, node.Content...)
+			case yaml.SequenceNode:
+				if n.Kind == 0 || n.Kind == yaml.SequenceNode {
+					n.Kind = yaml.SequenceNode
 				} else {
 					return fail(fmt.Errorf("mapping contains multiple kinds: had %v, now also %v", kindStr(n), kindStr(node)))
 				}
```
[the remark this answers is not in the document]
dagood: Added a test case that shows that this isn't real behavior.

internal/pipelineymlgen/eval.go
```diff
@@ -574,18 +610,16 @@ func (e *EvalState) evalRangeResult(r *evalRange) (*yaml.Node, error) {
...
-		return seq, nil
+		return e.mergeRangeItems(items)
 	}
 
 	// Handle slices and arrays via reflection to support typed slices (e.g. []string).
 	rv := reflect.ValueOf(r.collection)
 	if rv.Kind() == reflect.Interface || rv.Kind() == reflect.Ptr {
 		if rv.IsNil() {
-			return seq, nil
+			return &yaml.Node{Kind: yaml.MappingNode}, nil
```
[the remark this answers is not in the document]
dagood: Fixed, for consistency. However, this isn't reachable when using pipelineymlgen templates, so no test added.

internal/pipelineymlgen/eval.go
```diff
@@ -136,6 +136,11 @@ func (e *EvalState) eval(orig *yaml.Node) (any, error) {
 		if err != nil {
 			return fail(fmt.Errorf("converting doc content result to YAML node: %w", err))
 		}
+		if resultNode == nil {
+			// All content dissolved (e.g. all conditions false).
```
michelle-clayton-work: For my own learning can you explain what this means a little more and provide an example of when resultNode would be nil?

dagood: Sure. This PR in general lets more methods return nil to be able to inline ranges properly, and that's why the way this handling is being changed in various places in general.

In this `yaml.DocumentNode` case, that means a `nil` made it all the way to the top level--it's the only thing in the evaluated gen.yml doc at all. It's meant for this kind of test (which actually isn't part of the suite yet):

```yml
${ inlinerange "v" (list) }:
  ${ .v }: true
# Golden: {}
```

This helps the developer using pipelineymlgen by making it so they don't have to separately handle the 0 case vs the 1-to-many case. `{ "a": true, "b": true }` naturally goes to `{}` if we go from 2 to 0 elements.

Wrapping the `inlinerange` in a sequence, like this test, follows a different flow, so it's not a `nil` case:

```yml
- ${ inlinerange "v" (list) }:
    - ${ .v }
# Golden: []
```

This also helps out the developer.

---

Now... this other test that exercises this execution path (and is already in the suite!) actually doesn't make much sense:

```yml
# Check what happens if after evaluating, nothing is left.
# An empty mapping seems reasonable.
${ inlineif false }:
  failure
# Golden: {}
```

There's no basis for this resulting in `{}` rather than `[]`, or better, `null`. This behavior doesn't even help the developer: they can use `${ inlineelse }: {}` if they really want that behavior.

I think we ought to change this inlineif behavior to emit an error. That would actually help catch places where the "else" case was never intended to be reached! (Realistically: an `inlineif` `inlineelseif` chain without any valid reason for `inlineelse` to be reached.)

---

There are some clarity problems in the code at the moment IMO--it seems a little arbitrary whether certain behaviors are handled on the descent or the ascent back up the tree, for one. I also think that maybe using sentinel values rather than `nil` would be a better way to keep track of how certain outputs should be written back out as yml.

So far, I've been ok with leaving pipelineymlgen a little loose because we have reasonable test coverage of use cases that we care about, we don't intend for other teams to use this tool unless they expect breaking changes (microsoft/go-infra as a whole is still a v0 module), and dev iteration time is extremely quick compared to something like AzDO yml.

michelle-clayton-work (approved):

internal/pipelineymlgen/eval.go
dagood: > I think we ought to change this inlineif behavior to emit an error. That would actually help catch places where the "else" case was never intended to be reached! (Realistically: an `inlineif` `inlineelseif` chain without any valid reason for `inlineelse` to be reached.)

Added this change, and generalized a bit. E.g. an `inlinerange` can also lead to a descriptive error if it ends up not providing anything to a place where some value is needed:

```yml
example:
  title:
    ${ inlinerange (list) }:
      Never
# ...result to YAML node(s): inlinerange with an empty collection cannot appear as a mapping value
```

I'd say generally the intent is that when there's no clear answer about what type the yml node should be, it's an error. Maybe this is too strict, but it can be opened up later if something more useful comes up.

gdams (approved):

PR #465, opened 2026-04-08
Truncate test outputs

michelle-clayton-work:
The AzDO test viewer silently truncates test failure content at ~4000 characters without any indication. This makes it hard to tell when output has been cut off, especially when a long line happens to end at the boundary.

This PR adds a [truncateForAzDO](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) step in the json2junit converter that runs when a test case's output exceeds 4000 characters. When triggered, it:

Shortens known env var lines — PATH=, GOPATH=, GOROOT=, etc. are condensed to just the variable name and last path segment (e.g. PATH=".../bin"), since these are typically very long and not useful for diagnosing failures.
Truncates remaining long lines to 200 characters with [...] appended.
Hard-caps the total at 4000 characters.
Prepends a notice so it's immediately clear the output was intentionally truncated and where to find the full content.
Output under 4000 characters is left untouched.
- https://github.com/microsoft/go/issues/2219


michelle-clayton-work:
Azdo behavior not consistent so closing until we clarify the necessity of this

PR #467, opened 2026-04-09
Add update-dl subcommand to generate go-lab dl packages

gdams:
Add a new releasego subcommand "update-dl" that automates creating dl/msgo<version>/main.go files in the microsoft/go-lab repository for new Go releases. For each version, it fetches the source archive SHA256 hash from the GitHub release's assets.json and generates a dl package from a template. It then opens a PR with auto-merge enabled.

Also add the update-dl step to the release-go-images pipeline, running alongside publish-announcement with its own toggle parameter.

### DIFF
--- a/cmd/releasego/templates/dl.template.go.tmpl
+++ b/cmd/releasego/templates/dl.template.go.tmpl
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-// The msgo{{.Version}} command runs the go command from Microsoft Go {{.Version}}.
+// The msgo{{.Version}} command runs the go command from Microsoft build of Go {{.Version}}.
 //
 // To install, run:
 //
--- a/cmd/releasego/templates/dl.template.go.tmpl
+++ b/cmd/releasego/templates/dl.template.go.tmpl
@@ -8,7 +8,7 @@
 //	$ go install github.com/microsoft/go-lab/dl/msgo{{.Version}}@latest
 //	$ msgo{{.Version}} download
 //
-// And then use the go{{.Version}} command as if it were your normal go
+// And then use the msgo{{.Version}} command as if it were your normal go
 // command.
 //
 // See the release notes at https://github.com/microsoft/go/releases/tag/v{{.Version}}.
--- a/cmd/releasego/templates/dl.template.go.tmpl
+++ b/cmd/releasego/templates/dl.template.go.tmpl
@@ -1,0 +1,23 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+// The msgo{{.Version}} command runs the go command from Microsoft Go {{.Version}}.
+//
+// To install, run:
+//
+//	$ go install github.com/microsoft/go-lab/dl/msgo{{.Version}}@latest
+//	$ msgo{{.Version}} download
+//
+// And then use the go{{.Version}} command as if it were your normal go
+// command.
+//
+// See the release notes at https://github.com/microsoft/go/releases/tag/v{{.Version}}.
+//
+// File bugs at https://github.com/microsoft/go/issues/new.
+package main
+
+import "github.com/microsoft/go-lab/dl/internal/version"
+
+func main() {
+	version.Run("msgo{{.Version}}", "{{.SHA256}}")
+}
--- a/cmd/releasego/update-dl.go
+++ b/cmd/releasego/update-dl.go
@@ -6,8 +6,8 @@ package main
 import (
 	"bytes"
 	"context"
+	"crypto/sha256"
 	_ "embed"
-	"encoding/json"
 	"errors"
 	"flag"
 	"fmt"
@@ -21,7 +21,6 @@ import (
 	"time"
 
 	"github.com/google/go-github/v65/github"
-	"github.com/microsoft/go-infra/buildmodel/buildassets"
 	"github.com/microsoft/go-infra/githubutil"
 	"github.com/microsoft/go-infra/gitpr"
 	"github.com/microsoft/go-infra/goversion"
@@ -90,15 +89,15 @@ func updateDL(p subcmd.ParseFunc) error {
 		}
 
 		version := gv.Full()
-		log.Printf("Fetching SHA256 for version %s...\n", version)
-		sha256, err := fetchGoSrcSHA256(ctx, client, *goOrg, *goRepo, "v"+version)
+		log.Printf("Fetching assets.json SHA256 for version %s...\n", version)
+		assetsJSONSHA256, err := fetchAssetsJSONSHA256(ctx, client, *goOrg, *goRepo, "v"+version)
 		if err != nil {
-			return fmt.Errorf("error fetching SHA256 for version %s: %w", version, err)
+			return fmt.Errorf("error fetching assets.json SHA256 for version %s: %w", version, err)
 		}
-		log.Printf("SHA256 for %s: %s\n", version, sha256)
+		log.Printf("assets.json SHA256 for %s: %s\n", version, assetsJSONSHA256)
 		dlVersions = append(dlVersions, dlVersionData{
 			Version: version,
-			SHA256:  sha256,
+			SHA256:  assetsJSONSHA256,
 		})
 	}
 	if len(dlVersions) == 0 {
@@ -217,8 +216,10 @@ func updateDL(p subcmd.ParseFunc) error {
 	return nil
 }
 
-// fetchGoSrcSHA256 downloads the assets.json from a GitHub release and returns the GoSrcSHA256.
-func fetchGoSrcSHA256(ctx context.Context, client *github.Client, owner, repo, tag string) (string, error) {
+// fetchAssetsJSONSHA256 downloads the assets.json from a GitHub release and returns its content SHA256.
+// The dl tool in go-lab uses this hash to verify the integrity of assets.json before extracting
+// platform-specific download URLs and hashes from it.
+func fetchAssetsJSONSHA256(ctx context.Context, client *github.Client, owner, repo, tag string) (string, error) {
 	var release *github.RepositoryRelease
 	if err := githubutil.Retry(func() error {
 		var err error
@@ -257,16 +258,8 @@ func fetchGoSrcSHA256(ctx context.Context, client *github.Client, owner, repo, t
 		return "", fmt.Errorf("error reading assets.json: %w", err)
 	}
 
-	var assets buildassets.BuildAssets
-	if err := json.Unmarshal(data, &assets); err != nil {
-		return "", fmt.Errorf("error parsing assets.json: %w", err)
-	}
-
-	if assets.GoSrcSHA256 == "" {
-		return "", fmt.Errorf("GoSrcSHA256 is empty in assets.json for release %s", tag)
-	}
-
-	return assets.GoSrcSHA256, nil
+	hash := sha256.Sum256(data)
+	return fmt.Sprintf("%x", hash[:]), nil
 }
 
 // dlFilePath returns the path for a dl package's main.go file within the go-lab repo.
--- a/cmd/releasego/update-dl.go
+++ b/cmd/releasego/update-dl.go
@@ -29,7 +29,7 @@ import (
 	"github.com/microsoft/go-infra/subcmd"
 )
 
-var downloadHTTPClient = &http.Client{Timeout: 30 * time.Second}
+var downloadHTTPClient = http.Client{Timeout: 30 * time.Second}
 
 func init() {
 	subcommands = append(subcommands, subcmd.Option{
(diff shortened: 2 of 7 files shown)


cmd/releasego/update-dl.go
```diff
@@ -0,0 +1,303 @@
...
+	}
+
+	reviewAuther, err := gitHubReviewerAuthFlags.NewAuther()
+	if err != nil {
+		return fmt.Errorf("failed to get GitHub review auther: %w", err)
+	}
+
+	// Check that none of the files already exist.
+	for _, f := range files {
+		if _, err := githubutil.DownloadFile(ctx, client, org, repo, "main", f.path); err != nil {
```
dagood: This should use `githubutil.NewRefFS` so it works remote and locally with the core functionality split out.

eng/pipelines/release-go-images-pipeline.yml
```diff
@@ -196,6 +200,18 @@ extends:
                         -github-reviewer-app-private-key '$(BotAccount-review-bot-for-go-private-key)' 
                     displayName: 📰 Publish announcement details
 
+                - ${{ if eq(parameters.runUpdateDL, true) }}:
+                  - script: |
+                      releasego update-dl \
```
dagood, line 205: Needs to be tested to make sure it doesn't interrupt a release.

cmd/releasego/update-dl.go
```diff
@@ -0,0 +1,303 @@
...
+
+func updateDL(p subcmd.ParseFunc) error {
+	var releaseVersions string
+	var dryRun bool
+	var org string
+	var repo string
+	var goOrg string
+	var goRepo string
+
+	flag.StringVar(&releaseVersions, "versions", "", "Comma-separated list of version numbers for the Go release (e.g. 1.25.8-1,1.26.1-1).")
```
dagood: Should use `flag.String` with assignment to avoid duplicated/disconnected list of flag names. (Readability.)

cmd/releasego/update-dl.go
```diff
@@ -0,0 +1,303 @@
...
+	}
+	if len(goVersions) == 0 {
+		return fmt.Errorf("no valid versions found in -versions flag")
+	}
+	sort.Sort(goVersions)
+
+	// For each version, fetch the SHA256 from the release's assets.json.
+	dlVersions := make([]dlVersionData, 0, len(goVersions))
+	for _, gv := range goVersions {
+		version := gv.Full()
```
dagood, line 91: Could you do this as a single loop to avoid any unnecessary round-tripping (or highlight any intentional round-tripping)? I don't see any reason it has to be sorted right here.

cmd/releasego/update-dl.go
```diff
@@ -0,0 +1,303 @@
...
+			fmt.Println("=====")
+			if _, err := os.Stdout.Write(f.content); err != nil {
+				return err
+			}
+			fmt.Println("=====")
+		}
+		return nil
+	}
+
+	auther, err := gitHubAuthFlags.NewAuther()
```
dagood: This seems too late to make new flags?

cmd/releasego/update-dl.go
```diff
@@ -0,0 +1,303 @@
...
+		); err != nil {
+			return fmt.Errorf("error uploading file %s to branch %s: %w", f.path, branchName, err)
+		}
+	}
+
+	ownerRepo := fmt.Sprintf("%s/%s", org, repo)
+	prReq := prSet.CreateGitHubPR(
+		org,
+		title,
+		"**Automated Pull Request:** Adds dl packages for new Microsoft Go releases.\n"+
```
dagood:
```suggestion
		"**Automated Pull Request:** Adds dl packages for new Microsoft build of Go releases.\n"+
```

cmd/releasego/templates/dl.template.go.tmpl
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+// The msgo{{.Version}} command runs the go command from Microsoft Go {{.Version}}.
```
dagood:
```suggestion
// The msgo{{.Version}} command runs the go command from Microsoft build of Go {{.Version}}.
```

cmd/releasego/update-dl.go
```diff
@@ -0,0 +1,303 @@
...
+	var assets buildassets.BuildAssets
+	if err := json.Unmarshal(data, &assets); err != nil {
+		return "", fmt.Errorf("error parsing assets.json: %w", err)
+	}
+
+	if assets.GoSrcSHA256 == "" {
+		return "", fmt.Errorf("GoSrcSHA256 is empty in assets.json for release %s", tag)
+	}
+
+	return assets.GoSrcSHA256, nil
```
dagood: Why `GoSrcSHA256`?

cmd/releasego/update-dl.go
```diff
@@ -0,0 +1,303 @@
...
+}
+
+// dlFilePath returns the path for a dl package's main.go file within the go-lab repo.
+func dlFilePath(version string) string {
+	return fmt.Sprintf("dl/msgo%s/main.go", version)
+}
+
+// generateDLPRTitle generates a human-readable PR title for the dl update.
+func generateDLPRTitle(versions []string) string {
```
dagood, line 286: Could this use simpler/maintainable `strings.Join(x, ", ")` like the dev-facing create-release-day-issue does?

Can even avoid a matching noun: `"Update dl for Go " + strings.Join(versions, ",")`. Although for the tracking issue, we don't even care about matching:
* https://github.com/microsoft/go/issues/2129
* https://github.com/microsoft/go/issues/2175

dagood (changes requested):

cmd/releasego/update-dl.go
gdams: I'm not sure I understand what you're asking me?

eng/pipelines/release-go-images-pipeline.yml
gdams: https://dev.azure.com/dnceng/internal/_build/results?buildId=2949672

gdams: That created https://github.com/microsoft/go-lab/pull/326 which looks good.

cmd/releasego/update-dl.go
dagood: It doesn't make sense for the source tarball sha256 to appear in this code, why is it here?

I think this caused https://github.com/microsoft/go-lab/pull/328

gdams: yup I understand now, I've added a CI workflow to the repo too which will help catch this in the future: https://github.com/microsoft/go-lab/pull/330

gdams: updated to use the correct sha256

gdams:
@dagood I've updated and triggered https://dev.azure.com/dnceng/internal/_build/results?buildId=2950648.

This raised this PR: https://github.com/microsoft/go-lab/pull/331 (which now has the correct shasum and passed the new CI checks)

cmd/releasego/update-dl.go
```diff
@@ -0,0 +1,288 @@
...
+)
+
+func init() {
+	subcommands = append(subcommands, subcmd.Option{
+		Name:    "update-dl",
+		Summary: "Add dl packages to the go-lab repository for new Go releases.",
+		Description: `
+The update-dl command generates dl/msgo<version>/main.go files for each specified
+Go release version and creates a pull request on the go-lab repository. It fetches
+the source archive SHA256 hash from the GitHub release's assets.json file.
```
dagood, line 38: Not true now (hopefully? 😄)

cmd/releasego/update-dl.go
```diff
@@ -0,0 +1,288 @@
...
+			assetsAsset = release.Assets[i]
+			break
+		}
+	}
+	if assetsAsset == nil {
+		return "", fmt.Errorf("assets.json not found in release %s", tag)
+	}
+
+	// Download the assets.json file.
+	downloadClient := &http.Client{Timeout: 30 * time.Second}
```
dagood, line 245: The `http.Client` should be reused (global `var` is typical), and it seems odd to use `&` here rather than where the client is used.

cmd/releasego/update-dl.go
```diff
@@ -0,0 +1,288 @@
...
+
+// checkDLFilesNotExist checks that none of the given file paths already exist in the filesystem.
+// fsys can be a githubutil.NewRefFS (remote GitHub) or os.DirFS (local).
+func checkDLFilesNotExist(fsys githubutil.SimplifiedFS, paths []string) error {
+	for _, path := range paths {
+		_, err := fsys.ReadFile(path)
+		if err == nil {
+			return fmt.Errorf("file %s already exists", path)
+		}
+		if !errors.Is(err, githubutil.ErrFileNotExists) && !errors.Is(err, os.ErrNotExist) {
```
dagood, line 278: Hmm, seems like a `githubutil` bug if the fs impl doesn't return something that `Is` `fs.ErrNotExist`.

cmd/releasego/update-dl.go
```diff
@@ -0,0 +1,288 @@
...
+	prSet := gitpr.PRRefSet{Name: "main", Purpose: fmt.Sprintf("dl/%s/%d", slug, time.Now().Unix())}
+	branchName := prSet.PRBranch()
+
+	if err := githubutil.CreateBranch(ctx, client, *org, *repo, branchName, "main"); err != nil {
+		return fmt.Errorf("error creating branch %s: %w", branchName, err)
+	}
+
+	for _, f := range files {
+		log.Printf("Uploading %s...\n", f.path)
+		if err := githubutil.UploadFile(
```
dagood, line 183: Use the tree API to avoid noisy history and simplify retries, e.g.

https://github.com/microsoft/go-infra/blob/d1bd44ef34130d852455f81284b50a684d2b8b2a/cmd/releasego/update-azure-linux.go#L161-L192

cmd/releasego/update-dl.go
```diff
@@ -0,0 +1,288 @@
...
+		}
+	}
+
+	ownerRepo := fmt.Sprintf("%s/%s", *org, *repo)
+	prReq := prSet.CreateGitHubPR(
+		*org,
+		title,
+		"**Automated Pull Request:** Adds dl packages for new Microsoft build of Go releases.\n"+
+			"This PR was generated automatically using the [`update-dl.go`](https://github.com/microsoft/go-infra/blob/main/cmd/releasego/update-dl.go) script.")
+	createdPR, err := gitpr.PostGitHub(ownerRepo, prReq, auther)
```
dagood, line 203: The calls in this func need retry set up. (Maybe needs to move to use the GitHub client?)

internal/infrasort/goversions.go
```diff
@@ -14,6 +14,11 @@ type GoVersions []*goversion.GoVersion
 func (versions GoVersions) Len() int      { return len(versions) }
 func (versions GoVersions) Swap(i, j int) { versions[i], versions[j] = versions[j], versions[i] }
 func (versions GoVersions) Less(i, j int) bool {
+	return GoVersionDesc(versions[i], versions[j])
+}
+
+// GoVersionDesc reports whether a should sort before b in descending version order.
+func GoVersionDesc(a, b *goversion.GoVersion) bool {
```
dagood, line 21: Name isn't clear on first glance, should use the same word as `(versions GoVersions) Less` and https://pkg.go.dev/cmp#Less:

```suggestion
func GoVersionLess(a, b *goversion.GoVersion) bool {
```

(If this were `SortVersions` and `SortVersionsDesc` were an alternative implementation, that would make sense, but when comparing two args "desc" doesn't mean anything at first glance for me.)

cmd/releasego/update-dl.go
```diff
@@ -0,0 +1,288 @@
...
+//go:embed templates/dl.template.go.tmpl
+var dlTemplate string
+
+func updateDL(p subcmd.ParseFunc) error {
+	releaseVersions := flag.String("versions", "", "Comma-separated list of version numbers for the Go release (e.g. 1.25.8-1,1.26.1-1).")
+	dryRun := flag.Bool("n", false, "Enable dry run: do not push changes to GitHub.")
+	org := flag.String("org", "microsoft", "The GitHub organization for the go-lab repository.")
+	repo := flag.String("repo", "go-lab", "The GitHub repository name for the dl packages.")
+	goOrg := flag.String("go-org", "microsoft", "The GitHub organization for the Go releases repository.")
+	goRepo := flag.String("go-repo", "go", "The GitHub repository name for Go releases.")
```
dagood, line 59: For consistency (and ease of use), these should support `microsoft/go` and `microsoft/go-lab` format. E.g.:

https://github.com/microsoft/go-infra/blob/2861819e40b17b7775df11cab7da8df9f968f3db/cmd/releasego/create-release-day-issue.go#L58

dagood (commented):
A reliability concern with missing retry logic, but mostly cleanup suggestions.

gdams:
@dagood updated PTAL

cmd/releasego/update-dl.go
```diff
@@ -28,14 +29,16 @@ import (
 	"github.com/microsoft/go-infra/subcmd"
 )
 
+var downloadHTTPClient = &http.Client{Timeout: 30 * time.Second}
```
dagood: It's still odd to use `&` here.

dagood (approved):

qmuntal (approved):

PR #469, opened 2026-04-10
Add warning for truncation

michelle-clayton-work:
This pull request enhances the `json2junit` tool to improve the handling of large test output content, specifically for Azure DevOps (AzDO) users. It introduces a mechanism to prepend a warning message to test outputs that exceed AzDO's display limit, alerting users that their output may be truncated in the UI, while ensuring the full content is preserved in the raw output. Comprehensive unit tests are also added to verify this behavior.

**AzDO output warning improvements:**

* Added a constant (`azdoMaxChars`) to define the maximum number of characters AzDO displays before truncating test output, and a warning message (`azdoWarning`) to alert users when this limit is exceeded. Introduced the `warnLongContent` function to prepend this warning to long outputs without truncating the content. (`json2junit/json2junit.go`)
* Updated the processing logic for test suite and test case outputs to use `warnLongContent`, ensuring the warning is added when necessary for both system out and failure/skipped messages. (`json2junit/json2junit.go`) [[1]](diffhunk://#diff-a4075adc4a8a10a3be1eed359843c0565051a96ae37d2ffecc727f065cbe08f3L273-R293) [[2]](diffhunk://#diff-a4075adc4a8a10a3be1eed359843c0565051a96ae37d2ffecc727f065cbe08f3L295-R322)

**Testing and validation:**

* Added unit tests for `warnLongContent` to verify correct behavior for short, empty, boundary-length, and over-limit content, as well as to ensure that all original output is preserved and the warning is correctly prepended. (`json2junit/json2junit_test.go`)

### DIFF
--- a/json2junit/json2junit.go
+++ b/json2junit/json2junit.go
@@ -17,6 +17,26 @@ import (
 	"time"
 )
 
+const (
+	// azdoMaxChars is the approximate maximum number of characters the AzDO
+	// test viewer displays before silently truncating test failure content.
+	azdoMaxChars = 32000
+)
+
+var azdoWarning = []byte("[json2junit: Output is ~32000+ characters and may be truncated by AzDO. See raw test output for full content.]\n\n")
+
+// warnLongContent prepends a warning to test output that exceeds AzDO's display
+// limit. The content itself is not modified or truncated.
+func warnLongContent(content []byte) []byte {
+	if len(content) <= azdoMaxChars {
+		return content
+	}
+	result := make([]byte, 0, len(azdoWarning)+len(content))
+	result = append(result, azdoWarning...)
+	result = append(result, content...)
+	return result
+}
+
 // Options configures the conversion behavior.
 type Options struct {
 	// IncludePackageInTestName prefixes each test case name with its package path,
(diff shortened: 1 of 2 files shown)


michelle-clayton-work:
Message shows: 

- https://github.com/microsoft/go/pull/2228

image

michelle-clayton-work:
@copilot apply changes based on the comments in [this thread](https://github.com/microsoft/go-infra/pull/469#pullrequestreview-4091779992)

dagood (approved):
If we expected this to be around for a while, it might be a little better to make it configurable, with a default of not warning about this, so that it can be adjusted from microsoft/go (and anywhere else) when the limit is changed, and so it can be calibrated per AzDO instance.

But, as it stands, we expect to add this in as a best effort, and then remove it in 2-3 weeks when a better AzDO-side fix is deployed, so it seems good to me to keep it simple and avoid churning this PR. We can simply revert this code once AzDO has the fix.

dagood:
I opened an issue in the icebox to keep track of that:

* https://github.com/microsoft/go-lab/issues/325

PR #471, opened 2026-04-17
Enable whitespace linter with multi-if rule

michelle-clayton-work:
Enables the whitespace linter with multi-if: true so that multi-line if expressions are required to be followed by a newline (or comment). Also fixes pre-existing whitespace violations flagged by the default rules.

- https://github.com/microsoft/go-lab/issues/239

### DIFF
--- a/internal/pipelineymlgen/eval.go
+++ b/internal/pipelineymlgen/eval.go
@@ -117,7 +117,6 @@ func (e *EvalState) eval(orig *yaml.Node) (any, error) {
 	}
 
 	switch orig.Kind {
-
 	case yaml.DocumentNode:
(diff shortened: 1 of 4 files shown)


internal/pipelineymlgen/eval.go
```diff
@@ -288,7 +287,6 @@ func (e *EvalState) evalExpressionScalar(node *yaml.Node) (any, error) {
 			}
 
 			switch result := result.(type) {
-
```
dagood, line 291: Interesting--the intent here was for readability (when a newline is necessary between cases, also include a newline before the first case so it doesn't get lost in noise), but I don't know if that's really necessary.

dagood (approved):

PR #472, opened 2026-04-20
Add workflow to check line endings

michelle-clayton-work:
This pull request introduces a new GitHub Actions workflow to automatically check for files with CRLF or mixed line endings in the repository. This helps enforce consistent line endings as defined by the project's `.gitattributes` and prevents accidental commits with incorrect line endings.

Workflow addition:

* Added a reusable workflow `.github/workflows/check-line-endings.yml` that runs on pushes and pull requests to the `main` branch, checking for files with CRLF or mixed line endings and failing the workflow if any are found.

- https://github.com/microsoft/go-lab/issues/297


gdams:
Looks like it's already caught some

.github/workflows/check-line-endings.yml
```diff
@@ -0,0 +1,43 @@
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Check for CRLF or mixed line endings
+        shell: bash
+        run: |
+          set -euo pipefail
+          BAD=$(git ls-files --eol | awk '$1 ~ /(mixed|crlf)/ || $2 ~ /(mixed|crlf)/ {print $1, $2, $NF}')
```
dagood: Maybe it's better to lean on `.gitattributes`, in case there are exceptions where CRLF is expected to be checked in? (Windows test data in particular.) E.g.

```
git add --renormalize .
if ! git diff --cached --stat --exit-code; then ...
```

Perhaps also with a test that `.gitattributes` has applied something to the effect of `* text=auto eol=lf`. Copilot suggests a probe like this:

```
$ git check-attr text eol -- probe-file-intentionally-does-not-exit.xyz
probe-file-intentionally-does-not-exit.xyz: text: auto
probe-file-intentionally-does-not-exit.xyz: eol: unspecified
```

.github/workflows/check-line-endings.yml
```diff
@@ -0,0 +1,49 @@
...
+          set -euo pipefail
+          # Probe a nonexistent file to see what attributes apply by default.
+          # Without a default (e.g. `* text=auto`), the renormalize step would
+          # silently pass for files lacking an explicit attribute.
+          probe=$(git check-attr text eol -- probe-file-intentionally-does-not-exist.xyz)
+          echo "$probe"
+          if ! echo "$probe" | grep -qE '^[^:]+: text: (auto|set|lf|crlf)$'; then
+            echo "::error::.gitattributes is missing a default text rule (e.g. '* text=auto')."
+            exit 1
+          fi
```
dagood, line 39: Add additional check for `eol: lf`? (Which this repo doesn't specify yet.)

`* text=auto` is normally fine, but `* text=auto eol=lf` seems better to me for Go, to ensure tests and `go install`ed tools behave the same everywhere (more likely).

We did this in the winnative backend:

* https://github.com/microsoft/go-crypto-winnative/pull/155

dagood (approved):

PR #473, opened 2026-04-24
Add benchcheck command and reusable benchmark workflows

michelle-clayton-work:
Ports the benchcheck tool from openssl/.github/benchcheck and adds two reusable workflows plus a composite action so backend repos (go-crypto-winnative, go-crypto-darwin, openssl) can share a common benchmarking pipeline.


cmd/benchcheck/report.go
```diff
@@ -0,0 +1,155 @@
...
+	"strings"
+)
+
+func cmdReport(args []string) {
+	fs := flag.NewFlagSet("report", flag.ExitOnError)
+	runURL := fs.String("run-url", "", "URL of the workflow run (fallback for job links)")
+	jobURLsFile := fs.String("job-urls", "", "TSV file mapping artifact labels to job URLs")
+	overallResult := fs.String("overall-result", "success", "overall bench job result (success/failure)")
+	maxLen := fs.Int("max-len", 65536, "maximum output length for GitHub PR comment compatibility")
+	fs.Usage = func() {
```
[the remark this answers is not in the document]
dagood: Personally, I prefer (and think it's sufficient) to make the func that parses args simple enough that it's "trivially correct", which often means pulling out a lot of the logic into funcs that themselves can be tested more easily.

If the command itself were our product I'd want to run automated tests on the end-to-end command, but I don't think it's worth the maintenance effort/cost for these kinds of engineering tools.

.github/workflows/benchmark-conclusion.yml
```diff
@@ -0,0 +1,147 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+name: Benchmark conclusion (reusable)
```
dagood: Does this name show up anywhere? It just seems unusual to include `(reusable)` (also, in 2 of 3 files)

.github/workflows/benchmark-conclusion.yml
```diff
@@ -0,0 +1,147 @@
+# Copyright (c) Microsoft Corporation.
```
dagood: Maybe put these files in a dir rather than common prefix to avoid clutter over time?

```
.github/workflows/benchmark/conclusion.yml
.github/workflows/benchmark/run.yml
.github/workflows/benchmark/setup.yml
```

.github/workflows/benchmark-conclusion.yml
```diff
@@ -0,0 +1,147 @@
...
+# Licensed under the MIT License.
+
+name: Benchmark conclusion (reusable)
+
+# Aggregates per-cell benchmark artifacts (uploaded as benchstat-*) into a
```
dagood: Maybe the usage info deserves a readme to fully document the usage end to end? E.g. `.github/workflows/benchmark/README.md`

.github/workflows/benchmark-conclusion.yml
```diff
@@ -0,0 +1,147 @@
...
+        type: string
+        default: ''
+      bench-result:
+        description: 'needs.bench.result from the calling workflow.'
+        required: true
+        type: string
+      infra-ref:
+        description: |
+          Ref of microsoft/go-infra to check out for building benchcheck.
+          Should match the @ref the caller used on this workflow.
```
dagood: Call it `@<sha>` to match documentation above?


```suggestion
          Should match the @<sha> the caller used on this workflow.
```

.github/workflows/benchmark-conclusion.yml
```diff
@@ -0,0 +1,147 @@
...
+          # Fetch job URLs and find the step number for the Compare step to deep-link.
+          # Reusable-workflow jobs are named "<caller-job-name> / <inner-job-name>",
+          # so strip the trailing " / <inner>" before parsing the matrix portion.
+          gh api "repos/${{ github.repository }}/actions/runs/${{ github.run_id }}/jobs" \
+            --paginate --jq '
+              .jobs[] | select(.name | startswith("bench (")) |
+              (.name | sub(" / [^/]+$"; "") | ltrimstr("bench (") | rtrimstr(")") | split(", ")) as $p |
+              # Find the Compare step that actually ran (not skipped).
+              (first(.steps[] | select((.name | test("[Cc]ompare")) and .conclusion != "skipped") | .number) // null) as $step |
+              "\($p | join("-"))\t\(.html_url)\(if $step then "#step:\($step):1" else "" end)"' \
```
dagood: Could this reasonably be done by `benchcheck`? (And submitting the comment?)

.github/workflows/benchmark-conclusion.yml
```diff
@@ -0,0 +1,147 @@
...
+#   jobs:
+#     conclusion:
+#       needs: [setup, bench]
+#       if: always() && needs.setup.result == 'success'
+#       uses: microsoft/go-infra/.github/workflows/benchmark-conclusion.yml@<sha>
+#       with:
+#         head-ref: ${{ needs.setup.outputs.head-ref }}
+#         pr-number: ${{ needs.setup.outputs.pr-number }}
+#         bench-result: ${{ needs.bench.result }}
+#         infra-ref: <sha>  # same SHA as the @ pin above
```
dagood: I don't think this is maintainable--I don't think dependabot will update this, so it'll go out of sync immediately. (I don't have a good answer in mind, though...)

.github/workflows/benchmark-run.yml
```diff
@@ -0,0 +1,150 @@
...
+
+# Reusable workflow that runs a single benchmark matrix cell end-to-end:
+# checkout HEAD and BASE, run `go test -bench=.` against each, then run
+# benchstat + benchcheck check and upload the results as an artifact.
+#
+# Designed to be called from a per-backend `benchmark.yml` with a matrix
+# strategy. Pair with benchmark-setup.yml (head/base ref resolution) and
+# benchmark-conclusion.yml (PR comment) in this repo.
+
+name: Run benchmark cell
```
dagood: Why is this split into three yml files rather than one that contains multiple jobs?

dagood (commented):
A few questions/suggestions about the workflow files.

.github/workflows/benchmark-conclusion.yml
dagood: Copilot pointed out [`${{ github.workflow_sha }}`](https://docs.github.com/en/enterprise-cloud@latest/actions/reference/workflows-and-actions/contexts) that I think would work instead of passing anything explicitly, if it works how it sounds.

.github/workflows/benchmark-conclusion.yml
michelle-clayton-work: I don't think so. I think on one of the iterations copilot added it and I just missed it. I also don't think it make sense to include

.github/workflows/benchmark-run.yml
michelle-clayton-work: The motivation behind it was because there were some things that the callers of this workflow needed to do inbetween steps so it was in some ways easier to split up. But I think there is a strong (and prob correct) case to merge them. I think it will just change the way args are passed in. Let me combine and we can see if it looks better!

.github/workflows/benchmark-conclusion.yml
michelle-clayton-work: Consolidated the 3 yml's into one -- to me that made sense to delete the benchmark folder so I just made a benchmark-workflow.md file in the workflows folder, but let me know if somewhere else makes more sense. I was unsure since there isn't a workflows doc in the docs section

dagood: For me, this is new territory because of the yml sharing. Until now, every workflow in `.github/workflows/*.yml` is something that runs on the current repo, and each file is independent from the others. Apparently, reusable workflows [are supposed to be in this dir](https://docs.github.com/en/actions/concepts/workflows-and-actions/reusing-workflow-configurations#key-differences-between-reusable-workflows-and-composite-actions), though, so I guess this is just the way things are. 😄 

If there's only one yml file, I think keeping the documentation in the file is reasonable. If a markdown file is justified though, for `benchmark.yml` I'd prefer `benchmark.md` or `benchmark-README.md` because we're already in `.github/workflows/`.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,293 @@
...
+
+on:
+  workflow_call:
+    inputs:
+      matrix:
+        description: |
+          JSON object with an "include" array of bench-cell descriptors. See
+          the file header for the supported per-cell fields.
+        required: true
+        type: string
```
dagood, line 17: Ouch... Actions reusable workflows are far more limited than I expected. (Only "simple" types, pointing to this approach that leaves the user dealing with a stringy JSON.)

I feel like two reusable workflows would be a happy medium: `benchmark-setup.yml` creates `setup` and `conclude` (AFAIK order doesn't matter, only the `needs:`); `benchmark-job.yml` is used by a job that defines its own `matrix`, letting it use ordinary syntax.

Or: one workflow, but give it a big setup switch? Something like:

```yml
jobs:
  bench-setup:
    uses: microsoft/go-infra/.github/workflows/benchmark.yml
    with:
      create-setup-jobs: true
  bench:
    strategy:
      matrix:
        go-version: [1.25.x, 1.26.x]
        openssl-version: ["3.0.13", "3.1.5", "3.2.1", "3.3.1", "3.4.0", "3.5.0"]
        host: [ubuntu-22.04, ubuntu-24.04-arm]
        platform: [ubuntu]
        include:
          ...
    uses: microsoft/go-infra/.github/workflows/benchmark.yml
    with:
      (...ordinary other passthrough variables.)
```

But... I don't know if that *actually* works out nicely in Actions workflow yml.

I think since @qmuntal might be spending the most time interacting with this, and @gdams has more experience with Actions, might be worth deciding with them to find the most maintainable (partially least-duplication) thing with these constraints--in particular, will probably help me avoid going down the rabbit hole looking for better options. 😄 I think my standards are probably higher when I have to review/accept something on behalf of others who will have to maintain it.

gdams: I agree that the stringy JSON approach feels like it would be painful to maintain, especially once the matrix grows.

My preference would be the two-workflow split:

benchmark-setup.yml handles the shared setup/conclude jobs and any common orchestration.

benchmark-job.yml is called by the actual benchmark job, where the consuming workflow owns the matrix using normal GitHub Actions syntax.

That seems like the best compromise: we avoid forcing users into JSON-encoded matrices, keep the matrix readable/reviewable in the caller, and still centralise the duplicated benchmark logic.

I’m less keen on the single workflow with a create-setup-jobs switch. It might work, but it feels like we’d be encoding multiple modes into one reusable workflow, which could make the interface harder to reason about over time.

So I’d suggest we prototype the two-workflow version first and only fall back to the switch-based design if Actions makes the split awkward.

michelle-clayton-work: I tried to make a 2-file split work and don't think it can. A reusable-workflow call is a single opaque node in the caller's job graph and the caller can needs: the whole call, but jobs inside the called workflow can't needs: caller-side jobs. So if conclude lives inside the same file as setup, it has no way to wait for the caller's bench matrix — there's no name it can reference, and trying to express the dependency from the caller side (setup-and-conclude needs bench, bench needs setup-and-conclude) is a cycle. Putting conclude in benchmark-job.yml instead doesn't help either, since that file is matrix-instantiated N times by the caller, so you'd get N concludes each seeing only their own artifact rather than one aggregated comment.

michelle-clayton-work: So I think the best way forward would be to use a single file workflow. Similar reasons to why the 2-file split won't workm I think the switch won't work either. So I think we will just have to use a stringy JSON approach here

michelle-clayton-work: This is how it would have to be called https://github.com/microsoft/go-crypto-winnative/pull/166

.github/workflows/benchmark.md
```diff
@@ -0,0 +1,112 @@
+# Benchmark workflow
+
+[`benchmark.yml`](../.github/workflows/benchmark.yml) is a reusable workflow
+that runs Go benchmarks on a PR (or `workflow_dispatch` run), compares them to
```
dagood: Chop sentences rather than word wrap.

.github/workflows/benchmark.md
```diff
@@ -0,0 +1,112 @@
...
+   artifact.
+3. **`conclude`** — downloads all per-cell artifacts, builds a single
+   markdown report with [`cmd/benchcheck`](../cmd/benchcheck), posts (or
+   updates) the PR comment, and fails the job if any cell reported a
+   regression or test failure.
+
+The workflow self-resolves `microsoft/go-infra` at `${{ github.workflow_sha }}`
+when building `benchcheck`, so the binary always matches the workflow YAML
+the caller pinned to. Bumping the caller's `@<sha>` is sufficient — there is
+no separate `infra-ref` to keep in sync.
```
dagood: I don't think it's necessary to mention this--I don't think readers would expect to need to keep some second instance in sync.

.github/workflows/benchmark.md
```diff
@@ -0,0 +1,112 @@
...
+
+jobs:
+  benchmark:
+    uses: microsoft/go-infra/.github/workflows/benchmark.yml@<sha>
+    permissions:
+      actions: read
+      contents: read
+      pull-requests: write
+    with:
+      infra-ref: <sha>   # same SHA as the @ pin above
```
dagood: Out of date? Or did it regress in the implementation somehow?

.github/workflows/benchmark.md
```diff
@@ -0,0 +1,112 @@
...
+`benchcheck` binary. (`${{ github.workflow_sha }}` resolves to the caller's
+event SHA inside a reusable workflow, so it can't be used here.)
+
+If `microsoft/go-infra` ships release tags, add a `# vX.Y.Z` comment after
+each SHA so Dependabot can keep both pins up to date.
+
+## `matrix` cell fields
+
+`inputs.matrix` is parsed with `fromJSON` and used directly as the bench
+job's `strategy.matrix`, so any standard matrix shape works — top-level
```
dagood: While we're here:

```suggestion
job's `strategy.matrix`, so any [standard matrix shape](https://docs.github.com/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations) works — top-level
```

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,249 @@
...
+              core.setOutput('pr-number', String(context.payload.pull_request.number));
+            } else {
+              core.setOutput('head-ref', context.sha);
+              core.setOutput('base-ref', process.env.DISPATCH_BASE_REF || process.env.DEFAULT_BASE_REF);
+              core.setOutput('pr-number', '');
+            }
+
+  bench:
+    needs: setup
+    name: bench (${{ matrix.runs-on }}-go${{ matrix.go-version }}${{ matrix.fips-mode && format('-fips{0}', matrix.fips-mode) || '' }}${{ matrix.cgo-enabled && format('-cgo{0}', matrix.cgo-enabled) || '' }}${{ matrix.xcode-version && format('-xcode{0}', matrix.xcode-version) || '' }})
```
dagood, line 74: Does `${{ matrix.runs-on }}-go${{ matrix.go-version }}${{ matrix.fips-mode && format('-fips{0}', matrix.fips-mode) || '' }}${{ matrix.cgo-enabled && format('-cgo{0}', matrix.cgo-enabled) || '' }}${{ matrix.xcode-version && format('-xcode{0}', matrix.xcode-version) || '' }}` need to be identical to the artifact name later?

If so, that's a maintenance concern, but I fear Actions isn't powerful enough to resolve it and we'll have to just ignore it to be pragmatic...

If not, maybe format it with spaces to make it easier for humans?

.github/workflows/benchmark.md
michelle-clayton-work: Yeah this comment was stale -- will fix

.github/workflows/benchmark.md
michelle-clayton-work: I think this is correct. I think it was jsut the comments from above were stale

.github/workflows/benchmark.yml
michelle-clayton-work: Unfortunately I think it is just that Actions is not strong enough and can't deduplicate it. Adding comments to explain this

.github/workflows/benchmark.md
dagood: Why can't we use `${{ github.workflow_sha }}` instead anymore?
(And doesn't this make the workflow essentially unmaintainable with dependabot?)

dagood: (Needs to be addressed.)

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,255 @@
...
+          # Reusable-workflow matrix jobs are named "<caller-job-name> / bench (<label>)";
+          # strip the "<caller-job-name> / " prefix and "bench (...)" wrapper to get the
+          # label, which matches the artifact-name suffix used by the report.
+          gh api "repos/${{ github.repository }}/actions/runs/${{ github.run_id }}/jobs" \
+            --paginate --jq '
+              .jobs[] | select(.name | test(" / bench \\(")) |
+              (.name | sub("^.* / bench \\("; "") | sub("\\)$"; "")) as $label |
+              # Find the Compare step that actually ran (not skipped).
+              (first(.steps[] | select((.name | test("[Cc]ompare")) and .conclusion != "skipped") | .number) // null) as $step |
+              "\($label)\t\(.html_url)\(if $step then "#step:\($step):1" else "" end)"' \
```
dagood, line 195: Could this be done by `benchcheck`?

cmd/benchcheck/check.go
```diff
@@ -0,0 +1,361 @@
...
+		case strings.HasPrefix(line, "--- FAIL"):
+			lines = append(lines, prefix+line)
+		case strings.HasPrefix(line, "FAIL\t"):
+			lines = append(lines, prefix+line)
+		case !inCrash && isCrashLine(line):
+			inCrash = true
+			lines = append(lines, prefix+line)
+		case inCrash && line == "":
+			inCrash = false
+			lines = append(lines, "")
```
dagood, line 198: Consider using https://pkg.go.dev/cmd/test2json and parse the JSON to avoid implementing our own potentially more fragile parser?

dagood (changes requested):
The manually maintained ref needs to be addressed, the other two comments here could be split off as followup.

.github/workflows/benchmark.md
```diff
@@ -0,0 +1,88 @@
...
+          "runs-on": ["ubuntu-latest", "macos-15"],
+          "go-version": ["1.25", "1.26"],
+          "include": [
+            { "runs-on": "macos-15", "cgo-enabled": "1" }
+          ]
+        }
+```
+
+Pin the workflow to a SHA.
+The `benchcheck` binary is built from `microsoft/go-infra` at that same SHA (read from `job.workflow_sha`), so there is no second ref to keep in sync.
```
dagood: Mentioning a second ref here still doesn't make sense to me--I don't think it'll be understood without the context of the old approach already in mind. Instead, can inline the reason a second ref would have been necessary:

```suggestion
The `benchcheck` binary is built from `microsoft/go-infra` at that same SHA (read from `job.workflow_sha`), so there is never a mismatch in the shared workflow steps and `benchcheck`'s functionality.
```

.github/workflows/benchmark.md
```diff
@@ -0,0 +1,88 @@
...
+
+If `microsoft/go-infra` ships release tags, add a `# vX.Y.Z` comment after the SHA so Dependabot can keep the pin up to date.
```
dagood: I think we can simplify, because we do.


```suggestion
Add a `# vX.Y.Z` comment after the SHA so Dependabot can keep the pin up to date when a new go-infra tag is published.
```

.github/workflows/benchmark.md
```diff
@@ -0,0 +1,88 @@
...
+  contents: read
+  pull-requests: write
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  benchmark:
+    uses: microsoft/go-infra/.github/workflows/benchmark.yml@<sha>
```
dagood: Include the comment mentioned below.


```suggestion
    uses: microsoft/go-infra/.github/workflows/benchmark.yml@<sha> # <tag>
```

.github/workflows/benchmark.md
```diff
@@ -0,0 +1,88 @@
...
+
+Pin the workflow to a SHA.
+The `benchcheck` binary is built from `microsoft/go-infra` at that same SHA (read from `job.workflow_sha`), so there is no second ref to keep in sync.
+
+If `microsoft/go-infra` ships release tags, add a `# vX.Y.Z` comment after the SHA so Dependabot can keep the pin up to date.
+
+## `matrix` cell fields
+
+`inputs.matrix` is parsed with `fromJSON` and used directly as the bench job's `strategy.matrix`, so any [standard matrix shape](https://docs.github.com/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations) works — top-level arrays for cross-product, plus optional `include` / `exclude` to add or trim cells.
+Each cell entry supports:
```
dagood: Nit: it seems odd to call these "cells". It seems reasonable in the 2d case (this doc's example), maybe the 3d case, but openssl's configuration is currently 4 dimensional. How about just "entry"? Or each "matrix configuration"?

(Copilot isn't able to find official documentation that calls these cells.)

.github/workflows/benchmark.yml
michelle-clayton-work: Adding to https://github.com/microsoft/go-lab/issues/342

cmd/benchcheck/check.go
michelle-clayton-work: Also adding to https://github.com/microsoft/go-lab/issues/342!

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,252 @@
...
+        default: 'main'
+      dispatch-base-ref:
+        description: |
+          Override base ref. Caller should pass its workflow_dispatch
+          base-ref input here so manual runs can override the default.
+        required: false
+        type: string
+        default: ''
+      go-version:
+        description: 'Go toolchain version used to build benchcheck in the conclude job.'
```
dagood:
```suggestion
        description: 'Go toolchain version used to build benchcheck in the conclude job and to run the benchmarks.'
```

.github/workflows/benchmark.md
```diff
@@ -0,0 +1,88 @@
...
+Each cell's display name and uploaded artifact name are derived as `<runs-on>-go<go-version>[-fips<n>][-cgo<n>][-xcode<v>]`.
+The combination must be unique across the matrix.
+
+## Other inputs
+
+| input | default | meaning |
+|---|---|---|
+| `default-base-ref` | `main` | Base ref used for `workflow_dispatch` runs when `dispatch-base-ref` is empty. |
+| `dispatch-base-ref` | `''` | Caller should pass its `workflow_dispatch` `base-ref` input here. Ignored on `pull_request`. |
+| `go-version` | `1.25.x` | Go toolchain used to build `benchcheck` in the `conclude` job. |
```
dagood: I would omit this table and refer to the yml file to avoid duplicating.

dagood (commented):
Some doc suggestions.

.github/workflows/benchmark.md
michelle-clayton-work: updated to entry!

dagood (approved):

dagood:
Created https://github.com/microsoft/go-infra/releases/tag/v0.0.11 to include this workflow.

PR #480, opened 2026-05-12
Add retain build to release pipeline

michelle-clayton-work:
This pull request introduces a new mechanism to ensure that important Azure DevOps (AzDO) builds related to release pipelines are retained permanently for audit and traceability purposes. It adds a new `retain-build` subcommand to the `releasego` tool, and updates several pipeline templates to use this command for key builds in the release process.

Key changes include:

**New functionality for build retention:**

* Added a new `retain-build` subcommand to the `releasego` tool, which marks an AzDO build as "keep forever" using build ID and environment variables or explicit parameters. (`cmd/releasego/retain-build.go`)

**Pipeline and job template updates:**

* Introduced a new `retainBuild` parameter to the `eng/pipelines/jobs/releasego.yml` template, and added a step to call `releasego retain-build` when this parameter is set. [[1]](diffhunk://#diff-fdf75944d299752714b9cce983691a6ab73b6eb8dae5703d905c24771a83772fR21-R22) [[2]](diffhunk://#diff-fdf75944d299752714b9cce983691a6ab73b6eb8dae5703d905c24771a83772fR68-R72)
* Set `retainBuild: true` in both the main release build pipeline (`eng/pipelines/release-build-pipeline.yml`) and the innerloop release pipeline (`eng/pipelines/release-innerloop-pipeline.yml`) to enable permanent retention for orchestration builds. [[1]](diffhunk://#diff-350ed07f9d59b3673e906cd807e4a5adf4352ab11afacf9c66ed6215595235faR125) [[2]](diffhunk://#diff-0b9f5a909fdd2fbf456ff303f6e348ac6d99fe91a99556930853f09b13750e63R73)

**Ensuring retention of key inner builds:**

* Added steps in the release pipeline to explicitly retain the innerloop build and the main Microsoft Go build, since these builds are triggered by the release process but do not have direct context about being part of a release. [[1]](diffhunk://#diff-0b9f5a909fdd2fbf456ff303f6e348ac6d99fe91a99556930853f09b13750e63R107-R117) [[2]](diffhunk://#diff-5a6b652eb5c7ad18c67c616686458928bbd3f631ed906ea78ae0a1953eec81d4R100-R112)
* Added a step to retain the canary build in the release pipeline, again ensuring permanent retention for release-related builds.

- https://github.com/microsoft/go-lab/issues/59?reload=1?reload=1

### DIFF
--- a/cmd/releasego/retain-build.go
+++ b/cmd/releasego/retain-build.go
@@ -1,0 +1,85 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package main
+
+import (
+	"context"
+	"flag"
+	"log"
+	"os"
+	"strconv"
+
+	"github.com/microsoft/azure-devops-go-api/azuredevops/build"
+	"github.com/microsoft/go-infra/azdo"
+	"github.com/microsoft/go-infra/subcmd"
+)
+
+func init() {
+	subcommands = append(subcommands, subcmd.Option{
+		Name:    "retain-build",
+		Summary: "Mark an AzDO build to be retained forever (set keepForever=true).",
+		Description: `
+By default, retains the build that is currently running this command, using
+BUILD_BUILDID, SYSTEM_COLLECTIONURI, and SYSTEM_TEAMPROJECT from the environment.
+Pass -id, -org, or -proj to override.
+`,
+		Handle: handleRetainBuild,
+	})
+}
+
+func handleRetainBuild(p subcmd.ParseFunc) error {
+	id := flag.Int("id", 0, "The AzDO build ID to retain. Defaults to the current build (env BUILD_BUILDID).")
+	azdoFlags := azdo.BindClientFlags()
+
+	if err := p(); err != nil {
+		return err
+	}
+
+	// Apply env-var defaults so the typical in-pipeline call site only needs to pass -azdopat.
+	if *id == 0 {
+		if v := os.Getenv("BUILD_BUILDID"); v != "" {
+			parsed, err := strconv.Atoi(v)
+			if err != nil {
+				return err
+			}
+			*id = parsed
+		}
+	}
+	if *azdoFlags.Org == "" {
+		*azdoFlags.Org = azdo.GetEnvCollectionURI()
+	}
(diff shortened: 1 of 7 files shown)


cmd/releasego/retain-build.go
```diff
@@ -0,0 +1,95 @@
...
+	// Override the default "[Required]" usage strings: for retain-build, -org and
+	// -proj are env-defaulted (SYSTEM_COLLECTIONURI, SYSTEM_TEAMPROJECT) so that
+	// the typical in-pipeline call site only needs -azdopat. Without this, -h
+	// would mislead users about which flags must be passed explicitly.
+	if f := flag.CommandLine.Lookup("org"); f != nil {
+		f.Usage = "The AzDO organization URL. Defaults to env SYSTEM_COLLECTIONURI."
+	}
+	if f := flag.CommandLine.Lookup("proj"); f != nil {
+		f.Usage = "The AzDO project. Defaults to env SYSTEM_TEAMPROJECT."
+	}
```
dagood: I think this (and the assignment later) should be changed into something like a shared `azdo.BindClientFlagsWithEnvDefaults()`. Seems likely to be useful elsewhere and avoids the after-the-fact stringy lookup.

eng/pipelines/jobs/releasego.yml
```diff
@@ -18,6 +18,8 @@ parameters:
   releaseIssue: nil
   # A string of flags to pass "add-retry-instructions" for this job, or empty string to skip.
   retryInstructionsFlags: ''
+  # When true, mark this orchestration build to be retained forever for the release audit trail.
+  retainBuild: false
```
dagood: What is this parameter for?

If it should be here at all, I think defaulting to `true` would make more sense, to avoid accidentally missing retention in some cases.

michelle-clayton-work: I can remove! I guess my initial thought was to set it to false in case there is anything that runs this a lot (ie. we don't want to retain any runs) but looking at it now I think we can get rid of it. From the research I've done it doesn't seem like there are any high-frequency callers (and if we introduce any it's easy to add back!)

dagood (approved):

PR #485, opened 2026-05-18
telemetry: re-enable module counter, add CI detection

gdams:
Also implements `DetectCI` as a direct life and shift of https://github.com/dotnet/sdk/blob/main/src/Cli/Microsoft.DotNet.Cli.Definitions/Telemetry/CIEnvironmentDetectorForTelemetry.cs

### DIFF
--- a/telemetry/ci.go
+++ b/telemetry/ci.go
@@ -88,6 +88,10 @@ func envMap(env []string) map[string]string {
 }
 
 // isTrue reports whether the value is a common boolean-true string.
+// Matches the logic in dotnet/sdk EnvironmentVariableParser.ParseBool.
 func isTrue(v string) bool {
-	return v == "true" || v == "True" || v == "TRUE" || v == "1"
+	return v == "1" ||
+		strings.EqualFold(v, "true") ||
+		strings.EqualFold(v, "yes") ||
+		strings.EqualFold(v, "on")
 }
--- a/telemetry/ci.go
+++ b/telemetry/ci.go
@@ -1,0 +1,93 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package telemetry
+
+import "strings"
+
+// DetectCI inspects the given environment variables to determine which CI
+// system is in use. The env parameter should be in the same format as
+// os.Environ() (i.e. each entry is "KEY=VALUE"). It returns a short
+// identifier matching the go/ci counter values, or "" if no CI system is
+// detected.
+func DetectCI(env []string) string {
+	m := envMap(env)
+
+	// Azure Pipelines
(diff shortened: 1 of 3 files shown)


telemetry/config/config.json
```diff
@@ -68,7 +68,13 @@
 					"Name": "go/cgo:{enabled,disabled}"
 				},
 				{
-					"Name": "msgo/systemcrypto:{enabled,disabled}"
+					"Name": "go/ci:{appveyor,aws_codebuild,azdo,circleci,github,gitlab,google_cloud_build,jenkins,teamcity,travis}"
```
qmuntal:
```suggestion
					"Name": "msgo/ci:{appveyor,aws_codebuild,azdo,circleci,github,gitlab,google_cloud_build,jenkins,teamcity,travis}"
```

qmuntal (changes requested):

qmuntal (approved):

telemetry/ci.go
```diff
@@ -0,0 +1,97 @@
...
+
+	// Jenkins
+	// https://github.com/jenkinsci/jenkins/blob/master/core/src/main/resources/jenkins/model/CoreEnvironmentContributor/buildEnv.groovy
+	if m["BUILD_ID"] != "" && m["BUILD_URL"] != "" {
+		return "jenkins"
+	}
+
+	// Google Cloud Build
+	// https://cloud.google.com/build/docs/configuring-builds/substitute-variable-values#using_default_substitutions
+	if m["BUILD_ID"] != "" && m["PROJECT_ID"] != "" {
```
dagood, line 66: Ineffective later branch? Wouldn't it be better to report these as an individual bool best guess for each platform that can be refined independently, rather than short circuiting? Even if `BUILD_ID` didn't repeat, these are very generic, common names for env vars. I would say that if we *must* select one specific string, these cases arguably shouldn't even be included.

gdams: I was simply copying the dotnet implementation for now (https://github.com/dotnet/sdk/blob/main/src/Cli/Microsoft.DotNet.Cli.Definitions/Telemetry/CIEnvironmentDetectorForTelemetry.cs) We could further refine this if necessary?

dagood: I think I missed `&&` this morning. 😄 Technically, these being set in combination might be somewhat reliable. But I think it stands that a set of bools makes more sense than a single string. (This is very different logic compared to https://github.com/dotnet/sdk/blob/main/src/Cli/Microsoft.DotNet.Cli.Definitions/Telemetry/CIEnvironmentDetectorForTelemetry.cs.)

dagood: Simultaneous comment from my end. It's different because it just looks for any of these cases to return "in CI" or "not CI", not attempting to identify which one.

PR #486, opened 2026-05-18
Detect each CI system independently; counter for any CI

dagood:
* From https://github.com/microsoft/go-infra/pull/485#pullrequestreview-4311925349: detect each system independently without short-circuiting.
* Add `msgo/ci:ci` counter and add `ci` to the slice of detected CI when any CI is found.
  * This is more similar to the counter implemented by https://github.com/dotnet/sdk/blob/main/src/Cli/Microsoft.DotNet.Cli.Definitions/Telemetry/CIEnvironmentDetectorForTelemetry.cs
  * In theory, microsoft/go could handle the length check and adding `ci` part, and all this repo would need to do is change `config.json`. But, it seems reasonable enough to handle it here, without any clear reason not to.
  * My hypothesis is that this would potentially make it easier to assemble some telemetry queries on the resulting data, because the individual CI systems wouldn't have to be summed up.
    * Previously, there could be a check for "any msgo/ci counter", but if multiple are reported for a single toolset use, this might be harder.
    * I don't know if this is *actually* helpful--haven't worked with the Microsoft build of Go data--but seems intuitive to have in this situation.

### DIFF
--- a/telemetry/ci.go
+++ b/telemetry/ci.go
@@ -6,74 +6,81 @@ package telemetry
 import "strings"
 
 // DetectCI inspects the given environment variables to determine which CI
-// system is in use. The env parameter should be in the same format as
-// os.Environ() (i.e. each entry is "KEY=VALUE"). It returns a short
-// identifier matching the go/ci counter values, or "" if no CI system is
+// systems are in use. The env parameter should be in the same format as
+// os.Environ() (i.e. each entry is "KEY=VALUE"). It returns "ci" plus short
+// identifiers matching the go/ci counter values, or nil if no CI system is
 // detected.
-func DetectCI(env []string) string {
+func DetectCI(env []string) []string {
 	m := envMap(env)
+	var detected []string
+	addDetected := func(ci string) {
+		if len(detected) == 0 {
+			detected = append(detected, "ci")
+		}
+		detected = append(detected, ci)
+	}
 
 	// Azure Pipelines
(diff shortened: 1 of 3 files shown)


gdams (approved):

qmuntal:
I'm not sold on this change. Note that our .NET reference, `CIEnvironmentDetectorForTelemetry.cs`, is only used to detect if the toolchain is run on CI, not which CI is used. .NET documents that they upload telemetry that differentiate the CI system, but they still haven't implemented it, AFAIK.

So, back to this change, having `msgo/ci:ci,azdo` feels redundant, `ci` is already part of the name, and `msgo/ci:azdo,github` is confusing and would look like a coherency bug in the telemetry dashboard. I like the current approach of returning the most likely environment, which better represent reality.

I'm not opposed to having a `msgo/ci:ci` counter if the environment variable `CI` is detected, and there isn't any other match, or even when there are multiple possible matches instead of listing individual possibilities.

dagood:
> I like the current approach of returning the most likely environment, which better represent reality.

I didn't get this impression from the current implementation (e.g. surely `TEAMCITY_VERSION` is a more likely positive indicator than `BUILD_ID` and `BUILD_URL` both being set? 😄), but I do think that's reasonable if it's rearranged and documented.

dagood:
* https://github.com/microsoft/go-infra/pull/489

PR #488, opened 2026-05-19
telemetry/config: "go/platform/target/port" counter

qmuntal:
Upstream will remove the `go/platform/target/goos` and the `go/platform/target/goarch` counters in favor of the new `go/platform/target/port` counter, see https://go-review.googlesource.com/c/go/+/775981. The latter combines the two removed counter for the sake of consistency. This is good for us too, as our dashboard show some entries that missed either the port or the os.

The accepted values for the new counter are taken from https://go-review.googlesource.com/c/telemetry/+/776000.

While here, fix a typo in the `msgo/systemcrypto` counter.

### DIFF
--- a/telemetry/config/config.json
+++ b/telemetry/config/config.json
@@ -59,12 +59,6 @@
 					"Name": "go/subcommand:{build,install,run}"
 				},
 				{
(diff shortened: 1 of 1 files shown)


gdams (approved):

gdams:
One nit is that I prefer to keep the msgo specific counters at the bottom of the file for consistency but I'm not going to cry :D

gdams (approved):

PR #489, opened 2026-05-19
Clarify DetectCI precedence

dagood:
Clarify that DetectCI is intended to be approximate, where the most likely should win. Move `TEAMCITY_VERSION` to higher precedence because it includes the CI name.


qmuntal (approved):
Thanks!

gdams (approved):

PR #499, opened 2026-06-03
Add 'git go-patch shell' subcommand

michelle-clayton-work:
Implement the streamlined patch-development workflow from microsoft/go#750. 'git go-patch shell' opens an interactive shell in the submodule with a '(git-go-patch)' prompt prefix, optionally running apply/rebase first and extract on exit.

- -apply / -rebase / -no-extract flags
- prompt prefix preserves the user's existing prompt (pwsh/bash/zsh) and falls back to cmd.exe on Windows
- GIT_GO_PATCH_INTERACTIVE marks the session and warns on nested shells
- skips extract when a rebase/merge/cherry-pick/revert is in progress
- hint when -rebase is used without a prior apply

Also fix two pre-existing format-string bugs in warnIfOutsideSubmodule.

- https://github.com/microsoft/go/issues/750


cmd/git-go-patch/README.md
```diff
@@ -38,6 +38,33 @@ git go-patch -h
...
+### Streamlined workflow with `git go-patch shell`
+
+`git go-patch shell` automates the most common parts of the editing workflow. It opens an interactive shell whose working directory is already set to the submodule (so there's no need to `cd`), and it automatically runs `git go-patch extract` when you exit the shell. The shell's prompt is prefixed with `(git-go-patch)` to indicate you're in this mode.
+
+A typical session looks like this:
+
+```
+git go-patch shell -apply
+# edit commits in the submodule, e.g. with `git rebase -i` or `git go-patch rebase`
+exit
```
dagood: I think that it's important to support `exit 1` to quit the shell without rewriting the patch files. (This being `exit 0` implicitly.) This gives an escape hatch when the dev knows they messed up history really bad, and would e.g. create thousands of patch files if they let the tool do its thing. (Or: if it would overwrite existing patch file changes that would be difficult to recover.)

Maybe this is already implemented (haven't read through), but either way, would need to be documented. The output when entering the shell should also mention it IMO.

michelle-clayton-work: Makes sense. Point of clarification though -- since a shell exits with its last command's status  (a plain exit 1 can't be told apart from a last command that just happened to fail (failed grep, Ctrl-C, etc.)) so keying "discard" off non-zero would skip saving unexpectedly. Should we use a sentinel for "discard," or just prompt like "rewrite patch files?" after the shell closes?

michelle-clayton-work: discussed offline -- will either exit 0 or 1 instead (updated README with details!)

cmd/git-go-patch/README.md
```diff
@@ -38,6 +38,44 @@ git go-patch -h
...
+### Streamlined workflow with `git go-patch shell`
+
+`git go-patch shell` automates the most common parts of the editing workflow. It opens an interactive shell whose working directory is already set to the submodule (so there's no need to `cd`), and when you exit the shell with a status of 0 it automatically runs `git go-patch extract`. The shell's prompt is prefixed with `(git-go-patch)` to indicate you're in this mode.
+
+A typical session looks like this:
+
+```
+git go-patch shell -apply
+# edit commits in the submodule, e.g. with `git rebase -i` or `git go-patch rebase`
+exit
```
dagood: Include 0 so it works everywhere:

```suggestion
exit 0
```

cmd/git-go-patch/README.md
```diff
@@ -38,6 +38,44 @@ git go-patch -h
...
+# `git go-patch extract` runs automatically and rewrites the patch files
+```
+
+#### Saving vs. discarding your changes on exit
+
+When you leave the shell, `extract` runs only if the shell exits with status `0`. This gives you an escape hatch: if you've made a mess of the commit history and don't want the tool to rewrite (or overwrite) your patch files, exit with a non-zero status instead.
+
+* **Save** &mdash; exit with status `0`: `exit 0` (or just `exit` in PowerShell, where a plain `exit` is always `0`). `git go-patch extract` runs and rewrites the patch files.
+* **Discard** &mdash; exit with a non-zero status: `exit 1`. `extract` is skipped and your patch files are left untouched. If you actually wanted to save, run `git go-patch extract` yourself afterward.
```
dagood: * I think I'd prefer typical separators rather than em dash, but either way I don't think html escape is necessary?
* I don't think this needs to mention PowerShell behavior, this can be left to the next paragraph.
* Streamline in general for quick reference.

```suggestion
* `exit 0` to **save all changes you made in the submodule to the patch files**.
* `exit 1` to **discard your changes**.
```

cmd/git-go-patch/README.md
```diff
@@ -38,6 +38,44 @@ git go-patch -h
...
+
+Note the shell-specific behavior of a plain `exit`: in PowerShell it always reports status `0`, but in POSIX shells like bash and zsh it inherits the status of the last command you ran. So in bash/zsh, if your final command failed (for example a `grep` with no match), a plain `exit` can carry a non-zero status and skip `extract`. When in doubt, use `exit 0` explicitly to save.
```
dagood: More distinct formatting, streamline, and weave in the manual save note in a place that seems pretty reasonable to me:


```suggestion
> [!TIP]
> In PowerShell, a plain `exit` always reports status `0`, but in POSIX shells like bash and zsh `exit` inherits the status of the last command you ran.
> Due to this, we recommend always using `exit 0` so your result doesn't depend on the type of shell you are using at the time.
>
> If you use `exit` in a POSIX shell and accidentally discard your changes, you can run `git go-patch extract` to save the changes yourself.
```

cmd/git-go-patch/shell.go
```diff
@@ -0,0 +1,484 @@
...
+
+	// Mark the environment with the target submodule so scripts and a nested 'git go-patch shell' can
+	// detect this mode. The shell builders leave cmd.Env nil when they don't need to customize it,
+	// which means "inherit this process's environment"; start from os.Environ() in that case so
+	// appending doesn't drop it.
+	if cmd.Env == nil {
+		cmd.Env = os.Environ()
+	}
+	cmd.Env = append(cmd.Env, envInteractive+"="+dir)
+
```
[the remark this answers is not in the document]
michelle-clayton-work: Not needed -- when launching child processes via Go exec, duplicate env keys are resolved so the last value wins and the code appends the new value last so the override shoudl work

cmd/git-go-patch/shell.go
```diff
@@ -0,0 +1,484 @@
...
+			root = abs
+		}
+		args = append(args, "-c", root)
+	}
+
+	cmd := exec.Command(exe, args...)
+	cmd.Dir = dir
+	cmd.Stdin = os.Stdin
+	cmd.Env = append(os.Environ(), subcmd.EnvSuppressSuccessMessage+"=1")
+	return executil.Run(cmd)
```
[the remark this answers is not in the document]
michelle-clayton-work: same as above (not needed)

cmd/git-go-patch/shell.go
```diff
@@ -0,0 +1,484 @@
...
+	for name, content := range files {
+		if err := os.WriteFile(filepath.Join(tmpDir, name), []byte(content), 0o600); err != nil {
+			cleanup()
+			return nil, nil, fmt.Errorf("unable to write temp %s: %w", name, err)
+		}
+	}
+
+	cmd := exec.Command(shell, "-i")
+	cmd.Env = append(os.Environ(), "ZDOTDIR="+tmpDir)
+	return cmd, cleanup, nil
```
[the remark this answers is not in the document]
michelle-clayton-work: same as above

cmd/git-go-patch/shell.go
```diff
@@ -0,0 +1,484 @@
...
+	}
+	return cmdShellCmd(comspec), nil, nil
+}
+
+func cmdShellCmd(comspec string) *exec.Cmd {
+	cmd := exec.Command(comspec)
+	// cmd.exe builds its prompt from the PROMPT environment variable. "$P$G" is the default
+	// "current-path>" prompt, so prepend the indicator to it.
+	cmd.Env = append(os.Environ(), "PROMPT="+promptPrefix+"$P$G")
+	return cmd
```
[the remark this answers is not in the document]
michelle-clayton-work: same as above

cmd/git-go-patch/README.md
```diff
@@ -38,6 +38,44 @@ git go-patch -h
...
+
+Useful flags:
```
dagood: I think they're all useful. 😄 What I think might be excluded here in the future are rarely used flags:

```suggestion
Commonly used flags:
```

cmd/git-go-patch/README.md
```diff
@@ -38,6 +38,44 @@ git go-patch -h
...
+
+* `-apply`: run `git go-patch apply` before opening the shell.
+* `-rebase`: run `git go-patch rebase` (an interactive rebase) before opening the shell. Combine with `-apply` to apply and then immediately start a rebase. The rebase runs to completion first; if it stops (for example on a conflict or an `edit`/`break` step) the shell still opens so you can resolve it and run `git rebase --continue`.
```
dagood: Organize for readability, point to help with `--continue`:

```suggestion
* `-rebase`: run `git go-patch rebase` (an interactive rebase) before opening the shell.
  * Combine with `-apply` to apply and then immediately start a rebase.
  * The rebase runs to completion first. If it stops (for example on a conflict or an `edit`/`break` step) the shell still opens so you can resolve it and run `git rebase --continue`. See [Fix up patch files after a submodule update](#fix-up-patch-files-after-a-submodule-update) for rebase conflict resolution techniques.
```

(Step 6 of that section could be updated to mention "or exit from your `git go-patch shell` session".)

cmd/git-go-patch/README.md
```diff
@@ -38,6 +38,44 @@ git go-patch -h
 
 Sometimes you have to fix a bug in a patch file, add a new patch file, etc., and `apply`, `rebase`, and `extract` can help.
 
+### Streamlined workflow with `git go-patch shell`
+
+`git go-patch shell` automates the most common parts of the editing workflow. It opens an interactive shell whose working directory is already set to the submodule (so there's no need to `cd`), and when you exit the shell with a status of 0 it automatically runs `git go-patch extract`. The shell's prompt is prefixed with `(git-go-patch)` to indicate you're in this mode.
```
dagood: Throughout: break into lines to improve maintainability.

```suggestion
`git go-patch shell` automates the most common parts of the editing workflow.
It opens an interactive shell whose working directory is already set to the submodule (so there's no need to `cd`), and when you exit the shell with a status of 0 it automatically runs `git go-patch extract`.
The shell's prompt is prefixed with `(git-go-patch)` to indicate you're in this mode.
```

cmd/git-go-patch/shell.go
```diff
@@ -0,0 +1,484 @@
...
+	}
+	return strings.TrimSpace(string(out)) != "0", nil
+}
+
+// runSelf runs this same git-go-patch executable with the given subcommand, in the given working
+// directory, inheriting the current stdio. It forwards the "-c" repo root flag if it was set, so the
+// child command targets the same repository. The working directory is set to the submodule so
+// subcommands like "rebase" don't warn that the shell is being run from outside the submodule.
+func runSelf(dir, subcommand string) error {
+	exe, err := os.Executable()
```
dagood: Ah, I don't think it should be done this way--the call should be made within the Go code. If it's necessary to factor out `extract` so it's easier (or possible?) to run it with a Go function call, I think we should do that.

This avoids the need for things like `EnvSuppressSuccessMessage`, which don't seem like a good fit as a general `subcmd` feature. It also avoids process overhead and any challenges in figuring out the path.

cmd/git-go-patch/shell.go
```diff
@@ -0,0 +1,484 @@
...
+}
+
+// hasPatchCommitsToExtract reports whether the submodule at goDir has any commits on top of the base
+// recorded by 'apply' (read from statusFilePath). 'extract' formats one patch file per such commit
+// and overwrites the patches directory with the result, so when there are none it would delete every
+// existing patch file. Returning false lets the caller skip extraction to avoid that data loss.
+//
+// If the base can't be read (for example 'apply' was never run), it returns true so the caller
+// proceeds and lets 'extract' report any problem itself, matching the behavior without this guard.
+func hasPatchCommitsToExtract(statusFilePath, goDir string) (bool, error) {
```
dagood: I think this safety feature (and probably others) should be added to `extract` directly, not only apply when `shell` is used. I don't think an `extract` caller would ever use `extract` to eliminate all patches, and `extract` can emit a useful error message in this case.

(We could add a `-f` arg to do it anyway, but I don't think it's necessary.)

dagood (changes requested):
Initial review, focused on core mechanism and some doc notes.

cmd/git-go-patch/shell.go
michelle-clayton-work: I updated it so shell now calls apply/rebase/extract in-process and I reverted EnvSuppressSuccessMessage, but one open question on the empty-extract case: when you exit shell with no commits to extract, I currently catch errNoPatchCommits and print a short "nothing to save" note (exit 0).

Is that the best approach or do you think we hsould let the error surface? 

Basically the options:
Catch it (current): message is tailored to the shell workflow; exit stays 0 since nothing was wrong.
Let it surface: less coupling (shell doesn't need to know an extract-internal error) and extract already emits a useful message.

cmd/git-go-patch/shell.go
michelle-clayton-work: Moved this guard to extract as you suggested. There is another guard that  skps extract if there is a git operation in progress that is in shell only, because the documented git am conflict-fixup flow intentionally runs extract mid-operation and guarding it in extract would block that

cmd/git-go-patch/shell.go
michelle-clayton-work: Discussed offline: 

Went with letting it surface — dropped the catch in shell and the errNoPatchCommits sentinel, so extract owns the behavior and shell just fails with extract's message

cmd/git-go-patch/extract.go
dagood: Include the command--it's not worth including verbatim mode in `git go-patch shell` IMO, but if someone wanted to do this, point them to a command that includes the flag.


```suggestion
				"If this is a burden, consider using 'git go-patch extract -verbatim' to allow spurious changes but take ~%v.\n",
```

(Suggestion doesn't seem to be getting the original line right? Maybe because this is outside of the default scope of the PR diff. Odd.)

cmd/git-go-patch/extract.go
```diff
@@ -65,33 +65,54 @@ func handleExtract(p subcmd.ParseFunc) error {
...
+	// and the patches directory is overwritten with the result, so extracting with no such commits
+	// would delete every existing patch file. Refuse instead: no caller wants 'extract' to wipe all
+	// patches, and this catches mistakes like running it before 'apply' or before making any commits.
+	if since != "" {
+		hasCommits, err := submoduleHasPatchCommits(goDir, since)
+		if err != nil {
+			return err
+		}
+		if !hasCommits {
+			return fmt.Errorf("no commits to extract into patch files: the submodule has no commits in %v..HEAD; run 'apply' and make your edits first", since)
```
dagood: I don't think it's necessarily worth pointing out the "intentionally delete all patches" scenario directly, but I think the wording here should be looser to avoid potentially being misleading ("your edits" might be the removal of all the commits!):

```suggestion
			return fmt.Errorf("no commits to extract into patch files: the submodule has no commits in %v..HEAD; did you miss running 'apply' first?", since)
```

cmd/git-go-patch/README.md
```diff
@@ -38,6 +38,58 @@ git go-patch -h
...
+Commonly used flags:
+
+* `-apply`: run `git go-patch apply` before opening the shell.
+* `-rebase`: run `git go-patch rebase` (an interactive rebase) before opening the shell.
+  * Combine with `-apply` to apply and then immediately start a rebase.
+  * The rebase runs to completion first. If it stops (for example on a conflict or an `edit`/`break` step) the shell still opens so you can resolve it and run `git rebase --continue`. See [Fix up patch files after a submodule update](#fix-up-patch-files-after-a-submodule-update) for rebase conflict resolution techniques.
+* `-no-extract`: don't run `git go-patch extract` automatically on exit (run it yourself when ready).
+
+If a rebase, merge, cherry-pick, or revert is still in progress when you exit the shell, `extract` is skipped automatically so the patch files aren't rewritten from an incomplete state.
+`extract` is likewise skipped when the submodule has no commits on top of the recorded base — for example if you open a plain `git go-patch shell` (without `-apply`) on a submodule that has no patches applied — because extracting from an empty history would delete every patch file.
```
dagood: Maybe more scannable, and more concrete scenario:


```suggestion
There are some conditions where `extract` will be skipped regardless of exit code in order to fail safe.
This avoids writing patches from an incomplete state:

* A rebase, merge, cherry-pick, or revert is still in progress.
* The submodule has no commits on top of the recorded base.
  * For example, if you use `git submodule update` and then `git go-patch shell` (without `-apply`), extracting from the empty history would delete every patch file.
```

cmd/git-go-patch/README.md
```diff
@@ -38,6 +38,58 @@ git go-patch -h
...
+Pass `-apply` (or run `git go-patch apply` first) when you intend to edit and re-extract patches.
```
dagood: I don't think it's necessary to point this out again, the context doesn't seem different?

(It's also overly rigid--my shell might die (e.g. because my wsl/ssh session died) and I would not pass `-apply` so I can resume my work.)


```suggestion
```

cmd/git-go-patch/README.md
```diff
@@ -38,6 +38,58 @@ git go-patch -h
...
+
+The shell sets `GIT_GO_PATCH_INTERACTIVE` in its environment (to the submodule's path) so scripts (and an accidental nested `git go-patch shell` for the same submodule) can reliably detect the mode.
+The prompt is also prefixed on a best-effort basis, but prompt frameworks that re-render the prompt on every command (for example powerlevel10k or oh-my-posh transient prompts) may drop the `(git-go-patch)` prefix; the printed banner and the environment variable are the reliable indicators that you're in shell mode.
+
+On zsh, the shell launches with a temporary `ZDOTDIR` so it can load your config and then prepend the prompt prefix.
+If your `.zshrc` sources split configuration relative to `$ZDOTDIR` (for example `source $ZDOTDIR/aliases.zsh`), those lookups resolve against the temporary directory and are silently skipped inside the git-go-patch shell.
```
dagood: Hmm, this is making demands on how you configure your shell? It doesn't seem worth it to me... especially vs. the risky consequence of your environment working in a new way vs. the way it should.

Is there e.g. an escape hatch to turn this feature off if it's blocking your workflow? Is it easy to recognize that this is what's going wrong?

cmd/git-go-patch/README.md
```diff
@@ -38,6 +38,58 @@ git go-patch -h
...
+`extract` is likewise skipped when the submodule has no commits on top of the recorded base — for example if you open a plain `git go-patch shell` (without `-apply`) on a submodule that has no patches applied — because extracting from an empty history would delete every patch file.
+Pass `-apply` (or run `git go-patch apply` first) when you intend to edit and re-extract patches.
+
+The shell sets `GIT_GO_PATCH_INTERACTIVE` in its environment (to the submodule's path) so scripts (and an accidental nested `git go-patch shell` for the same submodule) can reliably detect the mode.
+The prompt is also prefixed on a best-effort basis, but prompt frameworks that re-render the prompt on every command (for example powerlevel10k or oh-my-posh transient prompts) may drop the `(git-go-patch)` prefix; the printed banner and the environment variable are the reliable indicators that you're in shell mode.
+
+On zsh, the shell launches with a temporary `ZDOTDIR` so it can load your config and then prepend the prompt prefix.
```
dagood: Organize for readability?


```suggestion
Some environment variables are set for the launched shell:

* `GIT_GO_PATCH_INTERACTIVE` is set to to the submodule's path.
  * Allows scripts to reliably detect the mode, and any accidental nested `git go-patch shell` for the same submodule to refuse to run.
* `PS1` is set to your existing `PS1` with the `(git-go-patch)` prefix inserted.

> [!NOTE]
> Prompt frameworks that re-render the prompt on every command (for example powerlevel10k or oh-my-posh transient prompts) may drop the prefix.
> The printed banner may be sufficient to recognize when you're in the shell.
> Or, consider integrating the `GIT_GO_PATCH_INTERACTIVE` environment variable into your custom dynamic prompt.

Additionally, if the active shell is zsh, the launched shell sets `ZDOTDIR` to a temporary directory that is set up to load your config and then prepend the prompt prefix.
```

cmd/git-go-patch/README.md
```diff
@@ -38,6 +38,58 @@ git go-patch -h
 
 Sometimes you have to fix a bug in a patch file, add a new patch file, etc., and `apply`, `rebase`, and `extract` can help.
 
+### Streamlined workflow with `git go-patch shell`
+
+`git go-patch shell` automates the most common parts of the editing workflow.
+It opens an interactive shell whose working directory is already set to the submodule (so there's no need to `cd`), and when you exit the shell with a status of 0 it automatically runs `git go-patch extract`.
+The shell's prompt is prefixed with `(git-go-patch)` to indicate you're in this mode.
```
dagood: This should maybe not be mentioned (the user will see it, or they won't, and it isn't necessary to know about it to use the tool properly), or it should include the "best effort" note here instead of below.

cmd/git-go-patch/shell.go
```diff
@@ -0,0 +1,417 @@
...
+		Description: `
+
+This command streamlines the common "apply, edit, extract" workflow by starting an interactive
+shell with its working directory set to the submodule, so no manual "cd" is necessary. The shell
+prompt is prefixed with "` + strings.TrimSpace(promptPrefix) + `" to make it clear the shell is in
+this special mode. When you exit the shell (for example by running "exit"), "git go-patch extract"
+runs automatically to rewrite the patch files based on the commits in the submodule.
+
+Inside the shell you can edit commits however you like, for example with "git rebase -i", and you
+can run "code ." to open an editor scoped to the submodule's history.
```
dagood: We shouldn't recommend `git rebase -i`, we already have two better utils for that (the `-rebase` flag and `git go-patch rebase`).

`code .` is an interesting hint, but seems more like a tip for the readme, or it could be super briefly in the banner that's printed when entering the new shell.

cmd/git-go-patch/shell.go
```diff
@@ -0,0 +1,417 @@
...
+Use "exit 0" to save the changes you made in the submodule to the patch files, or "exit 1" to
+discard them and leave the patch files untouched. We recommend always using "exit 0" explicitly: in
+PowerShell a plain "exit" always reports status 0, but in bash or zsh a plain "exit" inherits the
+status of the last command you ran. If you discard by accident, run "git go-patch extract" yourself
+to save the changes.
+
+If a rebase, merge, cherry-pick, or revert is still in progress when you exit the shell, "extract"
+is skipped to avoid rewriting the patch files from an incomplete state. "extract" is also skipped if
+the submodule has no commits on top of the recorded base (for example if you didn't pass "-apply"
+and no patches are applied), since extracting from an empty history would delete every patch file.
```
dagood: This is pretty long and repeats a few things as well as repeating what's in (or should be in) the flag docs. I think it's best to have a tight description here of how the command is intended to work and be used:

```suggestion
This command streamlines the common "apply, edit, extract" workflow by starting an interactive
shell with its working directory set to the submodule, so no manual "cd" is necessary.

When the shell exits successfully (when you execute "exit 0"), the "extract" subcommand runs
automatically. When the shell exits with non-zero exit code (for example when you execute "exit 1"),
the "extract" subcommand is not run. This allows you to skip extraction.

The "extract" subcommand is also skipped in situations where it is very likely undesirable:
- If a rebase, merge, cherry-pick, or revert is still in progress.
- If the submodule has no commits on top of the recorded base.
```

(I think that if we want to fully document these commands inside the command help, we should change how we do the README to share the source of truth. (E.g. have a `go generate` that extracts the fully rendered help text and puts it into a markdown doc.) I think we'd probably want to have a "short help" and "long help" as well to avoid overwhelming users.)

cmd/git-go-patch/shell.go
```diff
@@ -0,0 +1,417 @@
...
+			// An interactive rebase exits non-zero when it stops for a conflict or an "edit"/"break"
+			// step. Warn, but still open the shell so the user can resolve it there.
+			fmt.Printf("\nWARNING: 'git go-patch rebase' exited with an error: %v\n", err)
+			fmt.Println("If it stopped for conflicts or an 'edit'/'break' step, resolve it in the shell and run 'git rebase --continue' (or 'git rebase --abort').")
+		}
+	}
+
+	if alreadyInShellFor(goDir) {
+		fmt.Println("\nWARNING: it looks like you're already inside a 'git go-patch shell' for this submodule (" + envInteractive + " is set).")
+		fmt.Println("Opening another one will nest shells; remember to 'exit' each one.")
```
dagood: Is there a valid reason to allow doing this? It seems like this could be a fail condition earlier in the process.

cmd/git-go-patch/shell.go
```diff
@@ -0,0 +1,417 @@
...
+	}
+
+	fmt.Printf("\nStarting an interactive shell in %#q.\n", goDir)
+	if *noExtract {
+		fmt.Println("Type 'exit' to leave the shell. 'git go-patch extract' will NOT run automatically; run it yourself when you're ready.")
```
dagood: Clarify that it isn't an immediate instruction:


```suggestion
		fmt.Println("When you're done with your changes, type 'exit' to leave the shell. 'git go-patch extract' will NOT run automatically; run it yourself when you're ready.")
```

cmd/git-go-patch/shell.go
```diff
@@ -0,0 +1,417 @@
...
+	} else {
+		fmt.Println("Use 'exit 0' to save your changes to the patch files, or 'exit 1' to discard them.")
+		fmt.Println("(In PowerShell a plain 'exit' is always 0; in bash/zsh it inherits the last command's status, so prefer 'exit 0'.)")
```
dagood: IMO it's more noise than the extra line is worth to justify it here every time (it's also a property of POSIX compliant shells generally, not just those two):

```suggestion
```

cmd/git-go-patch/shell.go
```diff
@@ -0,0 +1,417 @@
...
+	}
+	fmt.Println()
+
+	shellErr := runInteractiveShell(goDir)
+
+	extract, launchErr := shouldExtractPatch(shellErr, *noExtract)
+	if launchErr != nil {
+		// The shell never ran, so there's nothing to extract; surface the failure.
+		return fmt.Errorf("failed to run the interactive shell: %w", launchErr)
```
dagood: Is there a reason to give the err a detailed name here?


```suggestion
	extract, err := shouldExtractPatch(shellErr, *noExtract)
	if err != nil {
		// The shell never ran, so there's nothing to extract; surface the failure.
		return fmt.Errorf("failed to run the interactive shell: %w", err)
```

cmd/git-go-patch/shell.go
```diff
@@ -0,0 +1,417 @@
...
+			// The shell exited non-zero, which we treat as "discard": skip extract and leave the
+			// patch files alone. Tell the user how to extract manually in case they meant to save.
+			var exitErr *exec.ExitError
+			if errors.As(shellErr, &exitErr) {
+				fmt.Printf("\nThe shell exited with status %d, so 'extract' was skipped and your patch files were left untouched.\n", exitErr.ExitCode())
+			} else {
+				fmt.Println("\nThe shell exited abnormally, so 'extract' was skipped and your patch files were left untouched.")
+			}
+			fmt.Println("If you meant to save your changes, run 'git go-patch extract'.")
+		}
```
dagood: I don't think we need to bring intent into it, and we can also include a copy-pasteable reminder in both cases this way:


```suggestion
		}
		fmt.Println("To manually save your changes, run 'git go-patch extract'.")
```

cmd/git-go-patch/shell.go
```diff
@@ -0,0 +1,417 @@
...
+		if p == "" {
+			continue
+		}
+		if !filepath.IsAbs(p) {
+			p = filepath.Join(dir, p)
+		}
+		switch _, err := os.Stat(p); {
+		case err == nil:
+			return true, nil
+		case !os.IsNotExist(err):
```
dagood: > New code should use errors.Is(err, fs.ErrNotExist).

https://pkg.go.dev/os#IsNotExist

cmd/git-go-patch/shell.go
```diff
@@ -0,0 +1,417 @@
...
+			return false, fmt.Errorf("unable to run 'git rev-parse --git-path %s': %w", name, err)
+		}
+		p := strings.TrimSpace(string(out))
+		if p == "" {
+			continue
+		}
+		if !filepath.IsAbs(p) {
+			p = filepath.Join(dir, p)
+		}
+		switch _, err := os.Stat(p); {
```
dagood: Unusual to use a switch AFAIK, would be easier to follow (and to skim) standard `!= nil` and followup checks.

cmd/git-go-patch/shell.go
```diff
@@ -0,0 +1,417 @@
...
+func interactiveShellCmd() (cmd *exec.Cmd, cleanup func(), err error) {
+	if runtime.GOOS == "windows" {
+		return windowsShellCmd()
+	}
+	return unixShellCmd()
+}
+
+func windowsShellCmd() (*exec.Cmd, func(), error) {
+	// Prefer PowerShell, which is the common shell for Microsoft build of Go development. Fall back
+	// to cmd.exe if PowerShell isn't available.
```
dagood: Is it possible(/reasonable) to detect the currently running shell? That definitely seems like what the expected behavior would be.

`pwsh` is also used on Linux, for example.

cmd/git-go-patch/shell.go
```diff
@@ -0,0 +1,417 @@
...
+		Handle: handleShell,
+	})
+}
+
+func handleShell(p subcmd.ParseFunc) error {
+	apply := flag.Bool("apply", false, "Run 'git go-patch apply' before opening the shell.")
+	rebase := flag.Bool("rebase", false, "Run 'git go-patch rebase' (interactive rebase) before opening the shell.")
+	noExtract := flag.Bool(
+		"no-extract", false,
+		"Don't automatically run 'git go-patch extract' when the shell exits.")
```
dagood, line 56: It's interesting that `-apply` and `-rebase` are opt-in, but `-no-extract` is opt out.

I think there's a good argument for changing to `-no-apply` (opt-out) as well.... If my repo is already set up with a manual `git go-patch apply`, running `apply` again should be a no-op, and it doesn't take long. If it isn't a no-op, then it at least fails instantly and I can slap `-no-apply` on the end of my command if that's what I want, or manually run `git go-patch apply -f` to clear any unwanted changes.

One potentially dangerous quirk: this adds a place where we depend strongly on `apply` detecting changes properly and refusing to run. If e.g. my session was lost and I'm creating a new `go-patch shell` to resume working, I will always have to pass `-no-apply` in that situation, and if I don't remember that, it runs this risk. 🤔

Interested what @qmuntal and @gdams think about this.

I think what's here now is the safest default and we could adjust it later, but if the `apply` change detection is working well for people (I tend to always pass `apply -f`, so I don't interact with that system 😅) then opt-out apply seems like a good idea.

---

I don't think there's a similar argument to swap to `-no-rebase`. When there are patch conflicts (the reason I most often use `git-go-patch` nowadays), I never want to rebase. If I'm exploring a change, I don't know which commit it should belong in, so it's way too early to do a rebase at the start, even if I do eventually do one. In the rare case that my explicit goal is to do some rebasing, the flag is nice to have.

cmd/git-go-patch/README.md
michelle-clayton-work: will delete

cmd/git-go-patch/README.md
michelle-clayton-work: Fair — the zsh ZDOTDIR rewrite is a lot of fragility for a cosmetic prefix. I will drop the prompt-prefix for zsh (and probably bash) and rely on GIT_GO_PATCH_INTERACTIVE + the entry banner as the reliable mode indicators

cmd/git-go-patch/shell.go
michelle-clayton-work: no -- rename is a good change

cmd/git-go-patch/shell.go
michelle-clayton-work: I don't think its super easy to do but I went with $SHELL + override rather than true parent-process detection. $SHELL is the configured shell (not necessarily the running one), but detecting the actual parent shell needs platform-specific syscalls or a new dep, plus ambiguous cases (parent is make/VS Code/ssh). I dont think Go's standard library has a cross-platform way to get the parent process's name/ $SHELL + -shell covers the real cases (pwsh-on-Linux, bash-on-Windows) without that complexity. I think the `-shell pwsh/bash/etc` seems like a good addition too to help with this issue.

cmd/git-go-patch/shell.go
michelle-clayton-work: I think the reasoning was that the detection is a heuristic, and it can false-positive. GIT_GO_PATCH_INTERACTIVE is set  in anything you launch from it — including code . → VS Code → VS Code's integrated terminal. If you open the repo in an editor from inside the shell and then run git go-patch shell in the editor's terminal, that's a legitimate action the heuristic would flag. 

But I think that maybe the warning is too light of a reaction so I think adding a -force flag would be a better option?

dagood: Interesting point... I'm fine leaving it as it is then. The lifecycle weirdness when a GUI app enters the picture and the process "detaches" does mean a hard block could be annoying to deal with when deep in a session.

dagood: I think `-force` would be ok, but that by itself is probably not clear when you might want to force the apply, or extract, etc., so maybe `-allow-self-nest` for this in particular? (I'm fine with or without making this a harder block though.)

cmd/git-go-patch/shell.go
dagood: Hmm, that is tricker than I thought! Agree a best effort with `SHELL` env + `-shell` flag is a good approach.

cmd/git-go-patch/shell.go
michelle-clayton-work: Applied your tighter version — dropped the per-flag repetition (the flag docs cover that) and the exit-detail that's in the README. 

Do you think the go generate is out of scope for this PR?

dagood: > Do you think the go generate is out of scope for this PR?

Definitely. 😄 There are a lot of existing things to convert and it would pile up this review a lot.

cmd/git-go-patch/shell.go
dagood: Ping @qmuntal and @gdams, when you use this tool I think you might agree swapping to `-no-apply` seems like a nice way to do it. But: we can make this change later, so not blocking merge on this thread.

cmd/git-go-patch/shell.go
```diff
@@ -0,0 +1,333 @@
...
+		}
+		return "cmd.exe"
+	}
+	if shell := os.Getenv("SHELL"); shell != "" {
+		return shell
+	}
+	return "/bin/sh"
+}
+
+type shellKindValue int
```
dagood, line 295: Just `shellKind` for idiomatic pseudo-enum naming:

```suggestion
type shellKind int
```

`func shellKind(shell string) shellKindValue` could be made more specific to free up the name:
`func parseShellBaseName(shell string) shellKind`

cmd/git-go-patch/shell.go
```diff
@@ -0,0 +1,333 @@
...
+			fmt.Printf("\nWARNING: 'git go-patch rebase' exited with an error: %v\n", err)
+			fmt.Println("If it stopped for conflicts or an 'edit'/'break' step, resolve it in the shell and run 'git rebase --continue' (or 'git rebase --abort').")
+		}
+	}
+
+	if alreadyInShellFor(goDir) && *allowSelfNest {
+		fmt.Println("\nWARNING: opening a nested 'git go-patch shell' for this submodule because -allow-self-nest was passed; remember to 'exit' each one.")
+	}
+
+	fmt.Printf("\nStarting an interactive shell in %#q.\n", goDir)
```
dagood, line 109: A nice line of 80 `=`s would be nice visual distinction for the beginning and closing banners. (Or a count matching shell width, but that seems to me like it'd probably be unnecessary complexity.)

This also helps distinguish at a glance when a shell was *really* launched and when `shell` returned an error message.

cmd/git-go-patch/shell.go
```diff
@@ -0,0 +1,333 @@
...
+			// An interactive rebase exits non-zero when it stops for a conflict or an "edit"/"break"
+			// step. Warn, but still open the shell so the user can resolve it there.
+			fmt.Printf("\nWARNING: 'git go-patch rebase' exited with an error: %v\n", err)
+			fmt.Println("If it stopped for conflicts or an 'edit'/'break' step, resolve it in the shell and run 'git rebase --continue' (or 'git rebase --abort').")
+		}
+	}
+
+	if alreadyInShellFor(goDir) && *allowSelfNest {
+		fmt.Println("\nWARNING: opening a nested 'git go-patch shell' for this submodule because -allow-self-nest was passed; remember to 'exit' each one.")
+	}
```
dagood, line 107: Checking if the submodule is currently in the middle of a rebase (etc.) and printing a quick note/warning would be nice for when you don't pass apply. I *think* it would be an easy addition because we already have `gitOperationInProgress`.

cmd/git-go-patch/shell.go
```diff
@@ -0,0 +1,333 @@
...
+	// the shell. Nesting has no real use (each exit would redundantly run 'extract'), so fail early
+	// rather than warn; -allow-self-nest overrides for the rare case the user really wants it.
+	if alreadyInShellFor(goDir) && !*allowSelfNest {
+		return fmt.Errorf("already inside a 'git go-patch shell' for this submodule (%s is set); "+
+			"exit it first, or pass -allow-self-nest to open a nested shell anyway", envInteractive)
+	}
+
+	if *apply {
+		if err := applyPatches(config, false, false, ""); err != nil {
+			return err
```
dagood, line 85: Error behavior for `-apply` is technically misleading:

```
dagood@dagood-3:~/git/go$ git go-patch shell -apply
[...]
Aborting: reapplying patches would discard changes in submodule. Use '-f' to proceed anyway
dagood@dagood-3:~/git/go$ git go-patch shell -apply -f
flag provided but not defined: -f
```

I think that adding a `-apply-f` flag, intercepting this type of error (adding a new error value if necessary) and mentioning the shell's flag instead would be the "maximum functionality" fix.

But: I think it'd be very reasonable to skip `-apply-f` and just intercept the error to tell the user to run `git go-patch apply -f`. I don't imagine a workflow would really benefit much from having a `shell` flag for this.

dagood (approved):
I just tried it out locally! Seems nice. I have a few comments, but at this point I'd be fine using this tool and anything else can be added as followup.

FWIW, for now, using `-shell pwsh` on Linux is a way to get the prompt to change even though we don't have a robust way to do it for posix shells yet:

```
(git-go-patch) PS /home/dagood/git/go/go> exit 1

The shell exited with status 1, so 'extract' was skipped and your patch files were left untouched.
To manually save your changes, run 'git go-patch extract'.

Success.
dagood@dagood-3:~/git/go$
```

michelle-clayton-work:
> I just tried it out locally! Seems nice. I have a few comments, but at this point I'd be fine using this tool and anything else can be added as followup.
> 
> FWIW, for now, using `-shell pwsh` on Linux is a way to get the prompt to change even though we don't have a robust way to do it for posix shells yet:
> 
> ```
> (git-go-patch) PS /home/dagood/git/go/go> exit 1
> 
> The shell exited with status 1, so 'extract' was skipped and your patch files were left untouched.
> To manually save your changes, run 'git go-patch extract'.
> 
> Success.
> dagood@dagood-3:~/git/go$
> ```

Merging! And created follow-up issue for cleaning up 
- https://github.com/microsoft/go-lab/issues/429

PR #502, opened 2026-06-04
Ship .asc signature alongside .sig

gdams:
Ship an `.asc` PGP signature file alongside the existing `.sig` for every archive. The `.sig` remains the default/primary extension; `.asc` is produced in parallel for compatibility with upstream Go tooling (e.g. official Dockerfiles).

See https://github.com/microsoft/go/issues/181

### Changes
- **buildassets.go**: Recognize both `.sig` and `.asc` when scanning build output directories (prefer `.sig`)
- **main.go**: Use `.sig` URLs for aka.ms links and release validation
- **dockerversions.go**: Update `PGPSignatureURL` doc comment
- **Tests**: Update test data and golden files

### Companion PR
- https://github.com/microsoft/go/pull/2348


buildmodel/buildassets/buildassets.go
```diff
@@ -246,9 +247,18 @@ func (b BuildResultsDirectoryInfo) CreateSummary() (*BuildAssets, error) {
...
+					sigAssociatedName = name
+					isSig = true
+					break
+				}
+			}
+			if isSig {
+				a := getOrCreateArch(sigAssociatedName)
 
 				a.PGPSignatureURL, err = getURL(e.Name())
 				if err != nil {
```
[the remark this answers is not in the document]
gdams: updated

cmd/releasego/testdata/Test_createLinkPairs/release_studio/aka.golden.txt
```diff
@@ -2,26 +2,26 @@ testing/go1.22.windows-amd64.zip             -> https://download.visualstudio.mi
 testing/go1.22.windows-amd64.zip.sha256      -> https://download.visualstudio.microsoft.com/download/pr/8330ba2b/41bc5/go1.23-90bcc55-20240626.1.windows-amd64.zip.sha256
 testing/go1.22.linux-amd64.tar.gz            -> https://download.visualstudio.microsoft.com/download/pr/8330ba2b/333aa/go1.23-90bcc55-20240626.1.linux-amd64.tar.gz
 testing/go1.22.linux-amd64.tar.gz.sha256     -> https://download.visualstudio.microsoft.com/download/pr/8330ba2b/87654/go1.23-90bcc55-20240626.1.linux-amd64.tar.gz.sha256
-testing/go1.22.linux-amd64.tar.gz.sig        -> https://download.visualstudio.microsoft.com/download/pr/8330ba2b/12345/go1.23-90bcc55-20240626.1.linux-amd64.tar.gz.sig
+testing/go1.22.linux-amd64.tar.gz.asc        -> https://download.visualstudio.microsoft.com/download/pr/8330ba2b/12345/go1.23-90bcc55-20240626.1.linux-amd64.tar.gz.asc
```
dagood: Missing legacy .sig?

(It's also not necessarily sufficient to create a `.tar.gz.sig` aka.ms URL that links to the same `.tar.gz.asc` file, because then people who do ordinary downloads will get a `.tar.gz.asc` file on disk they aren't prepared for. (Because it's a redirect, not opaque.) For clarity it might make the most sense to explicitly carry both files through the buildassets flow.)

dagood (changes requested):

buildmodel/buildassets/buildassets.go
```diff
@@ -117,9 +117,9 @@ func (b BuildAssets) GoVersion() *goversion.GoVersion {
 // archiving infra is stored in each release branch to make it local to the code it operates on and
 // less likely to unintentionally break, so some of that information is duplicated here.
 var (
-	archiveSuffixes = []string{".tar.gz", ".zip"}
-	checksumSuffix  = ".sha256"
-	signatureSuffix = ".sig"
+	archiveSuffixes   = []string{".tar.gz", ".zip"}
+	checksumSuffix    = ".sha256"
+	signatureSuffixes = []string{".sig", ".asc"}
```
dagood: Describe the order where declared so someone can't miss it if they were to add another or think that alphabetical order would be nice.

```suggestion
	archiveSuffixes = []string{".tar.gz", ".zip"}
	checksumSuffix  = ".sha256"
	// signatureSuffixes are the extensions used to store signatures, ordered by priority.
	signatureSuffixes = []string{".sig", ".asc"}
```

dagood (changes requested):
I'm not seeing how this ends up shipping both the .sig and .asc to aka.ms, or including both in the build assets JSON?

gdams:
> I'm not seeing how this ends up shipping both the .sig and .asc to aka.ms, or including both in the build assets JSON?

updated PTAL

buildmodel/buildassets/buildassets.go
```diff
@@ -119,7 +119,10 @@ func (b BuildAssets) GoVersion() *goversion.GoVersion {
 var (
 	archiveSuffixes = []string{".tar.gz", ".zip"}
 	checksumSuffix  = ".sha256"
-	signatureSuffix = ".sig"
+	// signatureSuffixes are the extensions used to store signatures, in alphabetical order. Both
+	// formats are produced and shipped: ".asc" is an ASCII-armored PGP signature and ".sig" is its
+	// binary equivalent.
```
dagood: Is it? The content should actually be the same...

cmd/releasego/main.go
```diff
@@ -47,6 +47,7 @@ func appendPathAndVerificationFilePaths(p []string, path string) []string {
 	p = append(p, path, path+".sha256")
 	if strings.HasSuffix(path, ".tar.gz") {
 		p = append(p, path+".sig")
+		p = append(p, path+".asc")
```
dagood, line 50: I'm not sure how this will interact with 1.25/1.26 (pre-asc)... this feeds into [some logic](https://github.com/microsoft/go-infra/blob/5930ff0b9943eda8c1e00a02ebeadca41d7a8607/cmd/releasego/repo-release.go#L159C6-L159C16) that I *think* expects all of these files to be present.

IIRC we've always had `.sig` (it just hasn't always been listed explicitly in the build assets json) so that's why there's some backfill but also an expectation that the file's always there.

I think the most reasonable way to get this done without much uncertainty would be to backport https://github.com/microsoft/go/pull/2348 to the active release branches.

dagood (changes requested):

cmd/releasego/main.go
gdams: agreed, I've raise backports to all release branches

buildmodel/buildassets/buildassets.go
gdams: Copilot hallucination, I've fixed the comments

dagood (approved):

PR #510, opened 2026-06-25
git-go-patch: address shell subcommand review follow-ups

michelle-clayton-work:
Implement deferred review suggestions from go-infra#499:

- Rename shellKindValue -> shellKind and shellKind() -> parseShellBaseName().
- Bracket the interactive session with 80-char '=' banners to distinguish a launched shell from an early error.
- Warn at startup if a rebase/merge/cherry-pick/revert is already in progress when neither -apply nor -rebase is passed.
- Intercept the dirty-submodule error from 'shell -apply' (new errSubmoduleDirty sentinel) and hint to run 'git go-patch apply -f'.
- https://github.com/microsoft/go/issues/2376

### DIFF
--- a/cmd/git-go-patch/shell.go
+++ b/cmd/git-go-patch/shell.go
@@ -94,7 +94,7 @@ func handleShell(p subcmd.ParseFunc) error {
 				return fmt.Errorf("%w\n\n"+
 					"'git go-patch shell -apply' will not discard these changes for you. Run "+
 					"'git go-patch apply -f' to discard them, or re-run 'git go-patch shell' without "+
-					"'-apply' to keep them.", err)
+					"'-apply' to keep them", err)
 			}
 			return err
 		}
@@ -129,6 +129,12 @@ func handleShell(p subcmd.ParseFunc) error {
 		}
 	}
 
+	shell := selectShell(*shellFlag)
+	resolvedShell, err := exec.LookPath(shell)
+	if err != nil {
+		return fmt.Errorf("failed to resolve interactive shell %q: %w", shell, err)
+	}
+
 	fmt.Println("\n" + sessionBanner)
 	fmt.Printf("Starting an interactive shell in %#q.\n", goDir)
 	fmt.Println("Edit the commits in the submodule however you like; run 'code .' to open an editor scoped to its history.")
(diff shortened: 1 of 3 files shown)


michelle-clayton-work:
@copilot address PR feedback and fix lint failures

gdams (approved):

cmd/git-go-patch/shell.go
```diff
@@ -115,14 +145,18 @@ func handleShell(p subcmd.ParseFunc) error {
...
 	fmt.Println()
 
-	shellErr := runInteractiveShell(goDir, *shellFlag)
+	shellErr := runInteractiveShell(goDir, resolvedShell)
 
 	extract, err := shouldExtractPatch(shellErr, *noExtract)
 	if err != nil {
-		// The shell never ran, so there's nothing to extract; surface the failure.
+		// The shell never ran, so there's nothing to extract; surface the failure without a closing
+		// banner so it's clear no interactive session actually started.
```
dagood: I'm not sure that's clear... I would interpret it as still being inside the interactive shell since there's only the one banner that was printed earlier. Matter of opinion, though, and rare that it'd come up.

cmd/git-go-patch/shell.go
```diff
@@ -115,14 +145,18 @@ func handleShell(p subcmd.ParseFunc) error {
...
 		return fmt.Errorf("failed to run the interactive shell: %w", err)
```
dagood, line 154: Not touched by this PR, but it would be more intuitive for this to be decided by `shouldExtractPatch`, rather than making an assumption here about what the `err` out of `shouldExtractPatch` means.

(When reading from the top, a straightforward `return err` here or something clearly saying "`shouldExtractPatch` told us that there was an error, so something fundamental went wrong and it's not possible to decide between "extract" and "don't extract"" would be more intuitive. This also leaves `shouldExtractPatch` open to include more error types in the future.)

dagood (approved):

cmd/git-go-patch/shell.go
michelle-clayton-work: clarified the comment!

dagood: The comment wasn't unclear, I was saying that I don't think this is clear:

`it's clear no interactive session actually started`

The first `===` banner did print, so I would expect that the interactive session did start, regardless of whether an error also happened. It would be clearer to me if `===` *were* printed in this case to make it clear that I'm not actually inside an interactive session.

cmd/git-go-patch/shell.go
dagood: To clarify: when I say "this" I mean the error message included in `("failed to run the interactive shell: %w", err)`. Based on the func name of `shouldExtractPatch` I think it would make more sense if `shouldExtractPatch` decided what the error text is.

dagood (approved):

PR #513, opened 2026-07-01
Benchmark workflow improvements

michelle-clayton-work:
Compare the least-performant base against the best-performing head using the confidence-interval bounds instead of the medians, so a regression is only flagged when the intervals do not overlap. Surface each sample's confidence-interval spread in the report so reviewers can see how noisy a benchmark is.

Also fail 'benchcheck report' on genuine result-file read errors instead of silently returning empty content.

- https://github.com/microsoft/go-lab/issues/342

### DIFF
--- a/cmd/benchcheck/report.go
+++ b/cmd/benchcheck/report.go
@@ -35,7 +35,7 @@ func cmdReport(args []string) {
 	}
 
 	resultsDir := fs.Arg(0)
-	jobURLs := loadJobURLs(*jobURLsFile)
+	jobURLs, readErr := loadJobURLs(*jobURLsFile)
 
 	var buf strings.Builder
 
@@ -82,8 +82,9 @@ func cmdReport(args []string) {
 			jobURL = *runURL
 		}
 		status := readJobStatus(filepath.Join(dir, "status.json"))
-		failures := readFileContent(filepath.Join(dir, "failures.txt"))
-		regressions := readFileContent(filepath.Join(dir, "regressions.txt"))
+		failures, failErr := readFileContent(filepath.Join(dir, "failures.txt"))
+		regressions, regErr := readFileContent(filepath.Join(dir, "regressions.txt"))
+		readErr = errors.Join(readErr, failErr, regErr)
 
 		if status.Failed() {
 			buf.WriteString("<details>\n")
@@ -122,6 +123,13 @@ func cmdReport(args []string) {
 	}
 
 	fmt.Print(output)
+
+	// A missing result file is expected (a job may not produce one); a genuine
+	// read error is not, and must fail the report rather than be swallowed.
+	if readErr != nil {
+		fmt.Fprintf(os.Stderr, "benchcheck report: %v\n", readErr)
+		os.Exit(1)
+	}
 }
 
 func readJobStatus(path string) Status {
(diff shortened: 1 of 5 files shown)


michelle-clayton-work:
@copilot address PR feedback

dagood:
Do you have an example of this running, for broad context?

cmd/benchcheck/report.go
```diff
@@ -141,34 +149,34 @@ func readJobStatus(path string) Status {
...
 		if len(parts) == 2 {
 			urls[parts[0]] = parts[1]
 		}
 	}
-	return urls
+	return urls, nil
 }
 
-func readFileContent(path string) string {
+func readFileContent(path string) (string, error) {
```
dagood: This name is a bit misleading--it's unclear why the caller wouldn't just use `os.ReadFile` themselves, and a future caller that takes `readFileContent` at face value could get in trouble based on those expectations.

Maybe `readTrimmedContentIfExists`?

Or move the `os.ErrNotExist` check to the caller, call it `readTrimmedFileContent`, and simplify this func by always returning `err`.

cmd/benchcheck/report.go
```diff
@@ -141,34 +149,34 @@ func readJobStatus(path string) Status {
 	return s
 }
 
-func loadJobURLs(path string) map[string]string {
+func loadJobURLs(path string) (map[string]string, error) {
```
dagood: I also wouldn't expect this to ignore missing files, but the name is a bit less standard so the expectations aren't so bad.

`readJobURLsFileIfExists` would be fine, I think.

It can also call `readTrimmedContentIfExists` to avoid the repeated error handling and trimming logic.

cmd/benchcheck/report.go
```diff
@@ -122,6 +123,13 @@ func cmdReport(args []string) {
 	}
 
 	fmt.Print(output)
+
+	// A missing result file is expected (a job may not produce one); a genuine
+	// read error is not, and must fail the report rather than be swallowed.
+	if readErr != nil {
```
dagood: Higher-level question: is it really a good design to intentionally be missing files? Could the system be set up to always create the expected file set, e.g. create an empty `failures.txt` file if there are no failures?

From a human side, it's simpler IMO to always create the files: an intentionally written empty file is more convincing than a file not existing. Also, does this process delete an existing `failures.txt` file if there are no failures, or does it leave it alone? You could end up thinking that you have errors even if you don't in that case. Always writing the file avoids that question and potential quirk of using deletion APIs rather than file-write APIs in some cases.

dagood (commented):
Thoughts on missing files.

michelle-clayton-work:
> Do you have an example of this running, for broad context?

Yes sorry forgot to link https://github.com/microsoft/go-crypto-winnative/pull/187

dagood (approved):
Nice, up to you if it makes sense to do anything now/later about the nonexistent file handling.

cmd/benchcheck/report.go
michelle-clayton-work: So in check.go it already always writes regressions.txt/failures.txt unconditionally, and the workflow pre-seeds status.json to benchmark_error:true. So empty files are the norm and nothing gets deleted. The only case where a file is genuinely absent is a job that crashes before check runs — which the pre-seeded status.json already reports as an error. The tolerant read just keeps report from failing on a partial/aborted artifact upload rather than being a design that relies on absence.

dagood: > The tolerant read just keeps report from failing on a partial/aborted artifact upload

Hmm, well, when would that happen? Why is it desirable to treat a partial/aborted artifact upload this way?

It's good that it doesn't rely on file absence, but that just means I'm not seeing a reason remaining to handle missing files in a special way rather than fail.

gdams (approved):

cmd/benchcheck/report.go
dagood: (Commenting on this as part of a review to make it more obvious.)

dagood (approved):
Still looks fine to me, just not sure the missing file handling approach makes sense.

cmd/benchcheck/report.go
michelle-clayton-work: You're right — the partial-upload rationale doesn't really hold up. I guess the only genuine reason report tolerated a missing .txt was the case where the workflow pre-seeds status.json to benchmark_error:true, but check is what writes the other two, so if a job died before check ran they'd be absent while report still had to render the error comment.

I changed it so that with absence no longer a normal state, report now reads those two files strictly — a missing file is a hard error — via a new readTrimmedFile. The tolerant readTrimmedContentIfExists stays only for the optional -job-urls input, whose absence really does just mean "not provided".

dagood (approved):

PR #516, opened 2026-07-06
benchcheck: parse go test -json instead of scanning text

michelle-clayton-work:
Replace the prefix-matching failure scanner with structured `test2json`
parsing, so failure detection no longer depends on matching `go test`
text output line-by-line.

## What changed

- **New `testjson.go`** — `parseTestJSON` reads `go test -json` output and:
  - detects build errors, failed tests, and crashes **structurally** from
    `fail` events (flushing the failing package/test's buffered output for
    context, dropping the `=== RUN`/`=== PAUSE`/`=== CONT` framing lines),
    instead of pattern-matching `# `, `--- FAIL`, `FAIL\t`, `SIGSEGV:`, and
    `panic:` prefixes;
  - reconstructs the plain benchmark text from the `output` events, which
    both `benchfmt` (regression parsing) and `benchstat` consume;
  - tolerates stray non-JSON lines.

- **`check.go`** — now takes `base.json`/`head.json`, parses each once for
  both failures and benchmarks, and adds `-o-base-text`/`-o-head-text` flags
  that write the reconstructed plain text for `benchstat`. Removed the old
  `extractFailures`, `isCrashLine`, `appendFailuresFromFile`, and
  `parseBenchmarksFromFile` helpers.

- **`benchmark.yml`** — the bench steps now emit `go test -json` to
  `base.json`/`head.json`. The "Compare and check" step runs `benchcheck`
  first (capturing its exit code) so it can produce the reconstructed
  `base.txt`/`head.txt`, then runs `benchstat` on them, then propagates the
  check exit code — so `benchstat.txt` is still produced even when a
  regression or failure makes `check` exit non-zero.

## Why

The text scanner was fragile: it relied on the exact shape of `go test`
output and on heuristics (e.g. cutting a crash trace at the first blank
line). Parsing the structured `test2json` stream makes failure detection
robust and lets us attribute output to the specific package or test that
failed.

## Testing

- `gofmt`, `go vet`, `go test ./cmd/benchcheck/` all clean; the text-scanner
  unit tests were replaced with `parseTestJSON` tests covering test
  failures, build failures, crashes, passing benchmark reconstruction, and
  non-JSON tolerance.
- End-to-end smoke test against crafted `test2json` inputs: reconstructed
  benchmark text is benchstat-ready, `failures.txt` captures the failing
  test and package summary, and `status.json` reflects the failure.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -142,24 +142,31 @@ jobs:
         shell: bash
         working-directory: base
         run: |
-          go test -run='^$' -bench=. -count=10 -benchmem -timeout 60m ./... 2>&1 | tee ../base.txt || true
+          go test -run='^$' -bench=. -count=10 -benchmem -timeout 60m -json ./... > ../base.json || true
 
       - name: Run benchmarks (head)
         shell: bash
         working-directory: head
         run: |
-          go test -run='^$' -bench=. -count=10 -benchmem -timeout 60m ./... 2>&1 | tee ../head.txt || true
+          go test -run='^$' -bench=. -count=10 -benchmem -timeout 60m -json ./... > ../head.json || true
 
       - name: "📊 Compare and check"
         shell: bash
         run: |
-          go -C _go-infra tool benchstat "$GITHUB_WORKSPACE/base.txt" "$GITHUB_WORKSPACE/head.txt" | tee benchstat.txt || true
           go build -C _go-infra/cmd/benchcheck -o "$PWD/benchcheck" .
+          # benchcheck reads the test2json output for structured failure
+          # detection and reconstructs the plain benchmark text that benchstat
+          # consumes. Run it first (capturing its exit code) so benchstat still
+          # runs even when a regression or failure makes check exit non-zero.
           ./benchcheck check \
+            -o-base-text=base.txt \
+            -o-head-text=head.txt \
             -o-regressions=regressions.txt \
             -o-failures=failures.txt \
             -o-status=status.json \
-            base.txt head.txt
+            base.json head.json && rc=0 || rc=$?
+          go -C _go-infra tool benchstat "$GITHUB_WORKSPACE/base.txt" "$GITHUB_WORKSPACE/head.txt" | tee benchstat.txt || true
+          exit $rc
 
       - name: Upload results
         if: always()
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -155,18 +155,26 @@ jobs:
         run: |
           go build -C _go-infra/cmd/benchcheck -o "$PWD/benchcheck" .
           # benchcheck reads the test2json output for structured failure
-          # detection and reconstructs the plain benchmark text that benchstat
-          # consumes. Run it first (capturing its exit code) so benchstat still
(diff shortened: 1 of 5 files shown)


gdams (approved):

.github/workflows/benchmark.yml
```diff
@@ -148,24 +142,31 @@ jobs:
...
           ./benchcheck check \
+            -o-base-text=base.txt \
+            -o-head-text=head.txt \
             -o-regressions=regressions.txt \
             -o-failures=failures.txt \
             -o-status=status.json \
-            base.txt head.txt
+            base.json head.json && rc=0 || rc=$?
+          go -C _go-infra tool benchstat "$GITHUB_WORKSPACE/base.txt" "$GITHUB_WORKSPACE/head.txt" | tee benchstat.txt || true
+          exit $rc
```
dagood: Instead of manual exit code handling, could you split this into two steps?

1. `./benchcheck check ...` gets built and run. The run has no special handling, and it determines the exit code since it's last.
3. `go -C _go-infra tool benchstat` runs. The step uses `if: always()`, so it runs even if the job has been marked as failed.

.github/workflows/benchmark.yml
```diff
@@ -148,24 +142,31 @@ jobs:
...
+          # runs even when a regression or failure makes check exit non-zero.
           ./benchcheck check \
+            -o-base-text=base.txt \
+            -o-head-text=head.txt \
             -o-regressions=regressions.txt \
             -o-failures=failures.txt \
             -o-status=status.json \
-            base.txt head.txt
+            base.json head.json && rc=0 || rc=$?
+          go -C _go-infra tool benchstat "$GITHUB_WORKSPACE/base.txt" "$GITHUB_WORKSPACE/head.txt" | tee benchstat.txt || true
```
dagood, line 177: Would it make more sense to run `benchstat` as part of `benchcheck`, avoiding some shuffling of data through bash (and through files)?

dagood (approved):
Looks good, a couple alternatives for a structural nitpick.

.github/workflows/benchmark.yml
michelle-clayton-work: done!

.github/workflows/benchmark.yml
michelle-clayton-work: Took another look into this and the pinned x/perf only exposes the deprecated benchstat library whose output differs from the cmd/benchstat CLI we use. Reproducing the current format in-process would mean coupling to benchstat's internals and I feel like that would be a lot of work for a smaller result so maybe I would lean towards keeping it as a CLI call

michelle-clayton-work:
updated the reference here to test the changes 
- https://github.com/microsoft/go-crypto-winnative/pull/187

.github/workflows/benchmark.yml
dagood: I'm not talking about it being in-process necessarily--the `benchcheck` tool could use `os/exec`. Maybe it's a wash, with the changes needed to do that. 🤷 

(If nobody will be attempting this sequence of commands locally, we also don't have that reason to try to make `benchcheck` handle more for them.)

dagood (approved):

PR #519, opened 2026-07-08
benchcheck: add job-urls subcommand to replace inline gh api --jq

michelle-clayton-work:
Move the benchmark workflow's job-URL lookup out of an inline
`gh api ... --jq` pipeline and into a tested `benchcheck job-urls`
subcommand.

## What changed

- **New `benchcheck job-urls [jobs.json]`** — reads the GitHub "list jobs
  for a workflow run" API response (from a file or stdin), and writes a TSV
  mapping each `bench (<label>)` matrix job to its job URL, deep-linked to the
  compare step when present. It handles the concatenated JSON pages that
  `gh api --paginate` produces. This is the same output the previous `--jq`
  program produced, and is consumed by `benchcheck report -job-urls`.

- **`benchmark.yml`** — the `conclude` job now builds `benchcheck` once, then
  runs `gh api ... --paginate > jobs.json` (gh still handles auth and
  pagination) piped through `benchcheck job-urls`, replacing the inline
  `--jq` filter. The prebuilt binary is reused by the report step.

## Why

The label/step-link extraction was an inline `jq` program embedded in YAML —
hard to read, hard to test, and easy to break. Moving it into `benchcheck`
makes it unit-testable and keeps the workflow step simple.

## Testing

`gofmt`, `go vet`, and `go test ./cmd/benchcheck/` all clean. New
`jobs_test.go` covers deep-linking to the compare step, a skipped compare
step (no deep link), ignoring non-benchmark jobs, multi-page (`--paginate`)
input, and empty input. Also smoke-tested the built binary against a sample
jobs response via stdin.

- https://github.com/microsoft/go-lab/issues/342

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -224,13 +224,14 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
-          # Reusable-workflow matrix jobs are named "<caller-job-name> / bench (<label>)".
-          # benchcheck job-urls extracts each label and its job URL (deep-linked
-          # to the compare step) from the jobs API response, matching the
-          # artifact-name suffix the report keys on.
-          gh api "repos/${{ github.repository }}/actions/runs/${{ github.run_id }}/jobs" \
-            --paginate > jobs.json || true
-          "$RUNNER_TEMP/benchcheck" job-urls jobs.json > job_urls.tsv || true
+          # benchcheck job-urls calls the GitHub jobs API (via gh, which handles
+          # auth and pagination) and writes a TSV mapping each matrix job
+          # "<caller-job-name> / bench (<label>)" to its URL, deep-linked to the
+          # compare step, matching the artifact-name suffix the report keys on.
+          "$RUNNER_TEMP/benchcheck" job-urls \
+            -repo "${{ github.repository }}" \
+            -id "${{ github.run_id }}" \
+            > job_urls.tsv
           cat job_urls.tsv
 
       - name: Build report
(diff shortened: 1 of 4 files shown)


.github/workflows/benchmark.yml
```diff
@@ -217,28 +217,26 @@ jobs:
...
           gh api "repos/${{ github.repository }}/actions/runs/${{ github.run_id }}/jobs" \
-            --paginate --jq '
-              .jobs[] | select(.name | test(" / bench \\(")) |
-              (.name | sub("^.* / bench \\("; "") | sub("\\)$"; "")) as $label |
-              # Find the Compare step that actually ran (not skipped).
-              (first(.steps[] | select((.name | test("[Cc]ompare")) and .conclusion != "skipped") | .number) // null) as $step |
-              "\($label)\t\(.html_url)\(if $step then "#step:\($step):1" else "" end)"' \
-            > job_urls.tsv || true
+            --paginate > jobs.json || true
+          "$RUNNER_TEMP/benchcheck" job-urls jobs.json > job_urls.tsv || true
```
dagood: Could this be `"$RUNNER_TEMP/benchcheck" job-urls -repo "${{ github.repository }}" -id "${{ github.run_id }}" > job_urls.tsv`, doing the `gh` internally?

This avoids the special exit code handling, which I think could cause an error to slip through if `job_urls.tsv` is partially written but the command fails. (A later step could fail due to the missing data in a confusing way.)

It also avoids keeping track of an intermediate file sitting around in the yml, and keeps the GitHub data model in Go code closer to the `gh` API call that it's used for.

michelle-clayton-work: Done updated it so job-urls now takes -repo/-id and runs gh api itself so the workflow step is just benchcheck job-urls  … > job_urls.tsv -- no intermediate jobs.json and no || true.

dagood (approved):

gdams (approved):

PR #523, opened 2026-07-14
Add pipeline self-description and adjust readmes

dagood:
* For https://github.com/microsoft/go-lab/issues/434
* Include small README adjustments: extra link for nav, point at style guide.
* While here, fix a few license headers that likely originally came from copying from a different repo.

### DIFF
--- a/eng/pipelines/release-build-pipeline.yml
+++ b/eng/pipelines/release-build-pipeline.yml
@@ -5,6 +5,14 @@
 
 # For info about runtime parameters, see https://github.com/microsoft/go-infra/blob/main/docs/pipeline-yml-style.md#runtime-parameters
 parameters:
+  - name: _info
+    displayName: >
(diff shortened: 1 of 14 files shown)


eng/pipelines/release-build-pipeline.yml
```diff
@@ -5,6 +5,14 @@
 
 # For info about runtime parameters, see https://github.com/microsoft/go-infra/blob/main/docs/pipeline-yml-style.md#runtime-parameters
 parameters:
+  - name: _info
+    displayName: >
+      Info: This pipeline releases a Microsoft build of Go by coordinating the microsoft/go PR, internal build, tags, GitHub release, aka.ms updates, and Go image update.
+    type: string
+    values:
+      - release-build-pipeline.yml
+    default: release-build-pipeline.yml
```
dagood, line 14: I didn't notice in my review that this was put directly into some templates rather than using the shared template. Reworking.

PR #527, opened 2026-07-20
godependabot: rename tool and generate YAML with Actions + gh aw support

gdams:
fixes: https://github.com/microsoft/go-lab/issues/475

Rename cmd/gomoddependabot to cmd/godependabot, since the tool now manages more than Go module ecosystems.

Changes:
- Emit the managed entries as readable multi-line YAML inside a marker-delimited block instead of single-line JSON, so the config is easy to read and hand-edited entries outside the block are preserved.
- Add a github-actions entry when a CodeQL workflow is detected, grouping github/codeql-action/* updates into a single PR.
- Add gh aw lock handling: when .github/workflows/*.lock.yml files exist, ignore github/gh-aw-actions* and exclude the lock files from updates. Both lines are annotated with "Managed by gh aw compile. Version-locked to the gh-aw compiler; do not bump." so their purpose is clear.
- Update .github/dependabot.yml to the new format and marker names.


gdams:
@michelle-clayton-work note that this update would allow to auto generate the dependabot files across all our repos with the required fix for the gh aw workflows.

karianna (approved):

gdams:
@copilot we also need to ignore `.github/workflows/copilot-setup-steps.yml` for agentic workflows. Add it as part of the ignore-path list

dagood:
> * hand-edited entries outside the block are preserved.

Was that not the case before? It was designed to preserve them, curious what happened here.

cmd/godependabot/main.go
```diff
@@ -0,0 +1,365 @@
...
+	dependabotPath := ".github/dependabot.yml"
+	oldData, err := os.ReadFile(dependabotPath)
+	if err != nil {
+		return fmt.Errorf("read existing yml: %v", err)
+	}
+	oldData = bytes.ReplaceAll(oldData, []byte{'\r', '\n'}, []byte{'\n'})
+
+	lines := strings.Split(string(oldData), "\n")
+
+	const beginGeneratedComment = "  # Following lines generated by godependabot; DO NOT EDIT."
```
dagood: Maybe use the full package path so it's easy to find when it's used on other repos?

cmd/godependabot/main.go
```diff
@@ -0,0 +1,365 @@
...
+			"Update .github/dependabot.yml based on the contents of the current Git repository. "+
+				"Adds an entry for every tracked Go module, and a github-actions entry when a CodeQL "+
+				"workflow is present. Updates are grouped so they're bundled as much as possible, reducing PR volume.")
+	}
+	flag.Parse()
+
+	dependabotPath := ".github/dependabot.yml"
+	oldData, err := os.ReadFile(dependabotPath)
+	if err != nil {
+		return fmt.Errorf("read existing yml: %v", err)
```
dagood, line 55: How about handling the missing file case with some initial info? Is that possible?

dagood (changes requested):
Binary file was committed.

cmd/godependabot/main.go
```diff
@@ -0,0 +1,365 @@
...
+		t := strings.TrimSpace(l)
+		isGHAWManaged := strings.HasPrefix(t, "- ") && (strings.Contains(t, lockWorkflowGlob) || strings.Contains(t, ghawDependencyName))
+		if isGHAWManaged {
+			line += " " + ghawManagedComment
+		}
+		newLines = append(newLines, line)
+	}
+	newLines = append(newLines, endGeneratedComment, "")
+
+	newData := []byte(strings.Join(append(lines, newLines...), "\n"))
```
dagood, line 162: It seems like this always moves the generated lines to the bottom. Is that the intent? (In that case, why bother with an "end" marker?)

cmd/godependabot/main.go
```diff
@@ -0,0 +1,365 @@
...
+		if !ok {
+			return false
+		}
+		var entry dependabotEntry
+		return json.Unmarshal([]byte(element), &entry) == nil
+	})
+
+	// Each generator inspects the repository and returns the dependabot entries
+	// it's responsible for. Add new ecosystem logic by appending a generator.
+	generators := []func(root string) ([]dependabotEntry, error){
```
dagood: Could change the signature to `func(entries []dependabotEntry, root string) ([]dependabotEntry, error)` and name them generally `appendGoModEntries` to follow that Go pattern.

cmd/godependabot/main.go
```diff
@@ -0,0 +1,365 @@
...
+		if line == "" {
+			continue
+		}
+		files = append(files, line)
+	}
+	return files, nil
+}
+
+func findGoModuleDirs(root string) ([]string, error) {
+	files, err := gitTrackedFiles(root)
```
dagood: Adjust API to avoid collecting data from Git multiple times?

cmd/godependabot/main.go
```diff
@@ -0,0 +1,365 @@
...
+		return nil, fmt.Errorf("find modules: %v", err)
+	}
+	var entries []dependabotEntry
+	for _, m := range modules {
+		entry := dependabotEntry{
+			PackageEcosystem: "gomod",
+			Directory:        m,
+			Groups: map[string]any{
+				"all": map[string]any{
+					"dependency-type": "production",
```
dagood: What is the intent? It [doesn't seem that dependabot supports the `dependency-type` option for go.mod currently](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#dependency-type-groups), but I also don't know what it means even with those docs. 😄

gdams: For what it's worth, this was already part of the existing generator. I'm not familiar with the syntax and would happily skip adding this config option.

cmd/godependabot/main.go
gdams: good point, I'll remove the end marker

gdams:
@dagood updated PTAL

cmd/godependabot/main.go
```diff
@@ -0,0 +1,363 @@
+// Copyright (c) Microsoft Corporation.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
```
dagood: Hmm, wrong license. Although it was wrong before, too, and it's also wrong in `.github/dependabot.yml`.

dagood (approved):

cmd/godependabot/main.go
```diff
@@ -0,0 +1,363 @@
...
+		// expects: "/" and "foo/bar" respectively.
+		line = "/" + line
+		if pre, ok := strings.CutSuffix(line, "/go.mod"); ok {
+			if pre == "" {
+				pre = "/"
+			}
+			dirs = append(dirs, pre)
+		}
+	}
+	sort.Strings(dirs)
```
michelle-clayton-work: just realized -- can we change this to slices.sort and then drop the sort import?

cmd/godependabot/main.go
dagood: I noticed that you actually wrote this part of the original generator in this PR:

* https://github.com/microsoft/go-infra/pull/439

(Not clear to me if there's a connection to update groups, or if the `*` was the only missing part.)

gdams: hmm let's remove it and see what happens

cmd/godependabot/main.go
gdams: Fixed

dagood (approved):

PR #533, opened 2026-08-06
Add reusable pull request classifier workflow

### LINKED ISSUE #767: Workflow to label PRs
Add a fork-safe GitHub Actions workflow that classifies each PR when it is opened, updated, reopened, or marked ready for review. Neither stage should check out or execute pull request code.

Size
Apply exactly one deterministic size label based on changed lines and file count:

- size:small: no more than 30 changed lines and 2 files.
- size:medium: no more than 100 changed lines and 5 files.
- size:large: no more than 300 changed lines and 10 files.
- size:xlarge: anything larger.

Kind
Apply every substantive kind that describes the PR:

- kind:code
- kind:docs
- kind:tests
- kind:examples
- kind:dependencies
- kind:ci

Mixed PRs may have multiple kind labels, such as kind:code and kind:tests. If the semantic classifier cannot read the PR or classify it confidently, preserve existing kind labels and apply failed-auto-classify; remove that marker after a later successful classification.

Area
Apply zero or more deterministic area labels based on changed paths:

- area:agent
- area:workflow
- area:message
- area:tool
- area:provider
- area:provider/a2a
- area:provider/agui
- area:provider/anthropic
- area:provider/copilot
- area:provider/foundry
- area:provider/gemini
- area:provider/openai
- area:provider/otel
- area:internal
- area:tooling
- area:examples
- area:github

Risk classification is tracked separately in #824.

michelle-clayton-work:
## Summary

- add a reusable Go workflow that deterministically applies `size:*` and caller-defined `area:*` labels
- add a reusable gh-aw workflow that uses AI to select and reconcile every applicable `kind:*` label
- provision all classifier-owned labels before the agent runs, including first-run and concurrent-creation handling
- support a single PR or a deterministic size/area backfill across all open PRs
- keep both stages API-only so they never check out or execute pull request code

## Kind labels

The agent may choose any substantive combination of:

- `kind:code`
- `kind:docs` (note: this includes comment only changes)
- `kind:tests`
- `kind:examples`
- `kind:dependencies`
- `kind:ci`

Safe outputs restrict it to those labels and allow removal only from the same set.

## Validation

- `go test -race ./cmd/prclassifier`
- `go vet ./cmd/prclassifier`
- `go test -run '^$' ./...`
- `gh aw compile classify-pull-request-kind --no-emit --validate`
- live GitHub API classification against PR #533
- [multi-kind end-to-end run 31417563708](https://github.com/microsoft/go-infra/actions/runs/31417563708) passed deterministic classification, agent execution, threat detection, safe outputs, and conclusion; it retained `kind:code` and added `kind:tests`, `kind:ci`, and `kind:docs`

Closes https://github.com/microsoft/agent-framework-go/issues/767


gdams:
I do wonder if some/all of this classification logic should be replaced with a reusable agentic workflow. It would probablty make it scale more easily and less prone to breaking?

michelle-clayton-work:
> I do wonder if some/all of this classification logic should be replaced with a reusable agentic workflow. It would probablty make it scale more easily and less prone to breaking?

I think so. I think area and size make sense to keep here but kind would be something that an agent should tackle. let me try to refactor and see if I can get that working

.github/workflows/classify-pull-request-kind.md
```diff
@@ -0,0 +1,80 @@
...
+timeout-minutes: 10
+---
+
+# Pull Request Kind Classifier
+
+Classify pull request `${{ inputs.pr_number }}` in `${{ github.repository }}` by its primary purpose and maintain exactly one `kind:*` label.
+
+## Allowed Kinds
+
+Choose exactly one:
```
gdams: I'm not sure choosing exactly one is correct here. E.g a PR might change both code and tests.

michelle-clayton-work: Good point. I changed kind classification to be multi-select: the agent now chooses every substantive kind, adds all missing kinds in one safe-output call, and removes only stale labels from the same allowlist.

The live E2E run on this PR passed and retained `kind:code` while adding `kind:tests`, `kind:ci`, and `kind:docs`: https://github.com/microsoft/go-infra/actions/runs/31417563708

michelle-clayton-work:
> > I do wonder if some/all of this classification logic should be replaced with a reusable agentic workflow. It would probablty make it scale more easily and less prone to breaking?
> 
> I think so. I think area and size make sense to keep here but kind would be something that an agent should tackle. let me try to refactor and see if I can get that working

Was able to refactor so now the kind tags are handled by an agent

gdams (approved):

.gitattributes
```diff
@@ -10,3 +10,5 @@
 # Use LF for all embedded markdown template files to avoid different results per
 # platform and mismatching tests.
 *.template.md text eol=lf
+
+.github/workflows/*.lock.yml linguist-generated=true merge=ours
```
dagood: Missing trailing newline.

Would be nice to also include a comment? (Especially if this is simply a standard suggestion for GHAW and there's an authoritative doc to point to.)

.github/workflows/classify-pull-request-kind.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Allowed Kinds
+
+Choose every kind represented by a substantive part of the change. Most PRs will have one or two kinds; a PR may have more.
+
+- `kind:dependencies`: dependency versions, module manifests, checksums, lock files, or dependency-bot maintenance change.
+- `kind:ci`: CI, build, repository automation, or GitHub workflows change.
+- `kind:docs`: documentation, guides, API documentation, or comments change substantively.
+- `kind:tests`: tests, fixtures, test data, or test infrastructure change.
+- `kind:examples`: samples, demos, or example registration change.
```
dagood: What does registration mean here?

.github/workflows/classify-pull-request-kind.md
```diff
@@ -0,0 +1,80 @@
...
+- `kind:code`: production behavior, public API, bug fixes, features, or refactors change.
+
+## Process
+
+1. Use GitHub tools to read the PR title, body, author, existing labels, changed files, and relevant diff patches.
+2. Determine the complete set of applicable kinds from the definitions above. Prefer the actual diff over the title. Do not add a kind merely because a generated file mentions it.
+3. Remove allowed `kind:*` labels that are currently present but no longer applicable in one `remove_labels` call. Do not remove labels outside the allowed set.
+4. Add all applicable labels that are not already present in one `add_labels` call, with a concise rationale and confidence for each.
```
dagood: Where does the rationale and confidence end up? Is that just for debugging when looking into a workflow execution?

.github/workflows/classify-pull-request-kind.md
```diff
@@ -0,0 +1,80 @@
...
+5. Do not add comments, reviews, or any labels outside the allowed set.
+6. If no label changes are needed, or if the PR cannot be read, use `noop` rather than making an unnecessary or speculative write.
```
dagood: If the PR can't be read or can't be classified for some other reason (like the agent refusing to touch it?), do we have a reasonable way to detect that something odd happened? Maybe a `failed-auto-classify` label would make sense?

.github/workflows/classify-pull-request.yml
```diff
@@ -0,0 +1,89 @@
...
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          repository: microsoft/go-infra
+          ref: ${{ job.workflow_sha }}
+          path: _go-infra
+
+      - name: Set up Go
+        uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
+        with:
+          go-version-file: _go-infra/go.mod
```
dagood: This leaves us with an old Go version, because we don't (and shouldn't) regularly update our `go.mod`'s `go` directive with patch updates. Using `go-version: '1.26'` is probably the best we can do for now.

.github/workflows/classify-pull-request.yml
```diff
@@ -0,0 +1,89 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+# Reusable workflow that applies size and repository-defined area labels
+# to pull requests. It reads PR metadata through the GitHub API and never checks
```
dagood, line 5: Maybe link to `cmd/prclassifier/classify.go` to show the thresholds the non-repository-defined size labels are based on?

cmd/prclassifier/classify.go
```diff
@@ -0,0 +1,130 @@
...
+	if fileCount == 0 {
+		fileCount = len(files)
+	}
+	filesTruncated := fileCount > len(files) || len(files) >= maxPullRequestFiles
+
+	size := "size:xlarge"
+	switch {
+	case changedLines <= 100 && fileCount <= 5:
+		size = "size:small"
+	case changedLines <= 300 && fileCount <= 10:
```
dagood: Have you run any stats on how many PRs would fall into which size categories? (Or something like that.)

I suspect that 100 lines && 5 files is quite big for being the smallest category, and it might group together quite substantial feature changes with one-line bugfixes.

cmd/prclassifier/config.go
```diff
@@ -0,0 +1,161 @@
...
+	if err := json.Unmarshal([]byte(value), &areas); err != nil {
+		return nil, fmt.Errorf("area-prefixes must be a JSON object: %w", err)
+	}
+	if areas == nil {
+		return nil, errors.New("area-prefixes must be a JSON object")
+	}
+	// Each PR may carry one size, one agent-selected kind, and every matching
+	// area label.
+	if len(areas)+2 > maxIssueLabels {
+		return nil, fmt.Errorf("area-prefixes defines too many labels: at most %d areas are allowed", maxIssueLabels-2)
```
dagood: If `maxIssueLabels` isn't the real limit, maybe define a new constant that includes the `-2` for simplicity? That would also be a place to describe why these limits are different, because that's not clear to me either. 😄

cmd/prclassifier/run_test.go
```diff
@@ -0,0 +1,208 @@
...
+	"bytes"
+	"context"
+	"errors"
+	"os"
+	"slices"
+	"strings"
+	"testing"
+)
+
+type fakeGitHub struct {
```
dagood, line 16: We use [`moq`](https://github.com/matryer/moq) elsewhere, could consider if it would work here:

https://github.com/microsoft/go-infra/blob/4d7b3d12f4a7adb007237d92f4f6fe3f9a59916b/cmd/releaseagent/internal/releasesteps/releasesteps.go#L17

(I'm not really sure that the way we have the dependency isolated in https://github.com/microsoft/go-infra/blob/main/eng/ci-tools/go.mod makes sense though, requiring extra steps to use actually install it to use it during code generation. I'll try a quick PR to make it easier to access.)

dagood (approved):
Some comments, but IMO none are blocking. :shipit:

.gitattributes
michelle-clayton-work: Fixed. I added the missing final newline and a comment linking to the gh-aw authoring guidance.

.github/workflows/classify-pull-request-kind.md
michelle-clayton-work: Meant catalog or index metadata used to make examples discoverable or runnable, not a special registration mechanism--I reworded it to make it more clear

.github/workflows/classify-pull-request-kind.md
michelle-clayton-work: That information would only have appeared in workflow artifacts or logs, not on the PR or labels. Since the E2E flow did not need it, I removed the rationale and confidence requirement.

.github/workflows/classify-pull-request-kind.md
michelle-clayton-work: Good idea -- added the label!

cmd/prclassifier/classify.go
michelle-clayton-work: I locally sampled 300 recently merged Agent Framework PRs. The old thresholds classified 210 of them (70%) as small. With the revised thresholds, the distribution is 107 small, 103 medium, 66 large, and 24 xlarge. I changed the boundaries to small ≤30 lines/2 files, medium ≤100/5, large ≤300/10, and xlarge above that. For the current 73 open PRs, this reduces the small bucket from 38 PRs to 2, making it a much stronger quick-review signal.

cmd/prclassifier/config.go
michelle-clayton-work: maxIssueLabels = 100 represents GitHub’s per-item limit. I added maxAreaLabels = 92, with a comment explaining that capacity is reserved for one size label, all six possible kind labels, and failed-auto-classify.

cmd/prclassifier/run_test.go
michelle-clayton-work: Sounds good. If this isn't blocking we can merge this PR without moq and refactor once that PR is in!

.github/workflows/classify-pull-request-kind.md
```diff
@@ -0,0 +1,83 @@
...
+   add-labels:
+      allowed:
+         - kind:code
+         - kind:docs
+         - kind:tests
+         - kind:examples
+         - kind:dependencies
+         - kind:ci
+         - failed-auto-classify
+      max: 6
```
dagood, line 45:
```suggestion
      # A PR may have any number of kinds, but a kind and the failure label are never applied simultaneously.
      max: 6
```

.github/workflows/classify-pull-request-kind.md
```diff
@@ -0,0 +1,83 @@
...
+   remove-labels:
+      allowed:
+         - kind:code
+         - kind:docs
+         - kind:tests
+         - kind:examples
+         - kind:dependencies
+         - kind:ci
+         - failed-auto-classify
+      max: 7
```
dagood, line 58:
```suggestion
      # A PR edited by a dev may have every kind and also the failure label.
      # The classifier may correctly decide a PR in that state has no kind, and all labels must be removed.
      max: 7
```

dagood (commented):
The 6 7 difference between the max added and removed is a bit mysterious... seems worth a note for whoever comes along next to update the list of kinds.

dagood (approved):

PR #535, opened 2026-08-07
Release UI for Go-images

michelle-clayton-work:
Create release UI that has go-images integrated.

Adds a local release-management UI for planning, running, monitoring, and resuming go-images releases through pipeline 1023. The Go server owns Azure access, validation, allowlisting, mirror checks, durable state, reconciliation, and SSE streaming. Embedded vanilla JavaScript calls local APIs and renders the dashboard, DAG progress, and two-step run confirmation. Supports normal, rollback, and test release modes with server-locked parameters.

Follow-up will refactor to more easily allow other processes to get integrated

image
image

- https://github.com/microsoft/go-lab/issues/480


cmd/releaseagent/internal/releaseui/web/go-images.html
```diff
@@ -0,0 +1,189 @@
+<!doctype html>
+<html lang="en">
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <title>Go Images Release · Microsoft Build of Go Release UI</title>
```
gdams:
```suggestion
  <title>Go Images Release · Microsoft build of Go Release UI</title>
```

cmd/releaseagent/internal/releaseui/web/go-images.html
```diff
@@ -0,0 +1,189 @@
...
+  <link rel="stylesheet" href="/assets/styles.css">
+  <script src="/assets/app.js" defer></script>
+</head>
+<body>
+  <header class="topbar">
+    <a class="brand" href="/" aria-label="Back to release dashboard">
+      <span class="brand-mark" aria-hidden="true">Go</span>
+      <span>
+        <strong>Release UI</strong>
+        <small>Microsoft Build of Go</small>
```
gdams:
```suggestion
        <small>Microsoft build of Go</small>
```

cmd/releaseagent/internal/releaseui/web/go-images.html
```diff
@@ -0,0 +1,189 @@
...
+    </a>
+    <a class="back-link" href="/">← All release processes</a>
+  </header>
+
+  <main class="release-main">
+    <section class="release-hero">
+      <div class="release-product-mark" aria-hidden="true">GI</div>
+      <div>
+        <p class="eyebrow">Go images · Pipeline 1023</p>
+        <h1>Release Microsoft Build of Go container images.</h1>
```
gdams:
```suggestion
        <h1>Release Microsoft build of Go container images.</h1>
```

michelle-clayton-work:
@copilot+gpt-5.6-sol address PR feedback and fix the linter

cmd/releaseagent/internal/releasesteps/releasesteps.go
```diff
@@ -9,19 +9,46 @@ import (
...
 // Input is the collection of inputs for a given release that don't change. They are provided once
 // by the release runner and stay the same upon retry.
 type Input struct {
 	// Versions is a list of versions to release.
 	Versions []string
 
+	// GoImagesReleaseMode and source fields bind a focused go-images plan to one allowlisted
+	// operation and the exact microsoft/main commit resolved by the server. SourceBuildID is set
+	// only for rollback and identifies the prior pipeline 1023 artifacts to republish.
+	GoImagesReleaseMode   GoImagesReleaseMode
```
dagood, line 57: The just-go-images release flow needs its own Input struct--there is a *lot* in here that doesn't apply. Other release types will have their own collections of inputs that don't change throughout each release process. Maybe some will be shared, and this could be done by reusing structs (or potentially embedding them).

cmd/releaseagent/internal/coordinator/step.go
```diff
@@ -20,7 +21,10 @@ type StepFunc func(ctx context.Context) error
 // Step represents a step in the release. Just enough information is represented in Step to allow
 // status to be reported, otherwise state is internal to Func.
 type Step struct {
-	// Name is the name of the step. It must be unique within the release step graph.
+	// ID is the stable, machine-readable identity of the step. It must be unique within the release
+	// step graph and must not change when display text changes.
+	ID string
+	// Name is the human-readable name of the step.
 	Name string
```
dagood, line 25: Why add `ID`? I think that for a system like this where the people who use the tool are the same people who maintain the tool, it's hard to justify fancy names and the maintenance that comes with them.

cmd/releaseagent/internal/coordinator/step.go
```diff
@@ -73,15 +80,113 @@ func NewIndicatorStep(name string, dependsOnAdditional ...*Step) *Step {
...
+		ID:        id,
 		Name:      name,
 		Timeout:   timeout,
 		Func:      f,
 		DependsOn: append(dependsOnAdditional, s),
 	}
 }
 
+// ValidateSteps checks that steps form a complete, executable directed acyclic graph.
+func ValidateSteps(steps []*Step) error {
```
dagood: Doesn't `TransitiveDependencies` already do this?

cmd/releaseagent/internal/azdopipeline/client.go
```diff
@@ -0,0 +1,380 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+// Package azdopipeline provides the narrow Azure DevOps Build API surface needed by release UI
+// workflows. It does not log tokens or request authorization headers.
+package azdopipeline
```
dagood, line 6: These client packages should be unnecessary, check out https://github.com/microsoft/go-infra/blob/main/cmd/releasego/build-pipeline.go for example.

cmd/releaseagent/README.md
```diff
@@ -1,10 +1,139 @@
...
 Subcommands:
 
-* `releaseagent run [...]` - Run the release agent. (Not yet implemented.)
-* `releaseagent write-mermaid-diagram` - Writes a mermaid diagram showing the steps and dependencies of the release process.
+* `releaseagent serve` starts the local release UI.
+* `releaseagent write-mermaid-diagram` writes a Mermaid diagram of the broader release process.
 
-See [ADR-0005 Use a release agent to coordinate releases](https://github.com/microsoft/go-lab/blob/main/docs/adr/0005-use-a-release-agent-to-coordinate-releases.md) for more information.
+The landing page is a release dashboard. It lists work tracked by the current durable session and a
+catalog of release processes. Go images is the only available process in this iteration; go-infra
```
dagood: Chop on sentence, not a column.

cmd/releaseagent/internal/releaseui/web/go-images.html
```diff
@@ -0,0 +1,189 @@
+<!doctype html>
+<html lang="en">
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <title>Go Images Release · Microsoft build of Go Release UI</title>
```
dagood, line 6: I don't think it makes sense to have separate HTML per type of queue. I think of the HTML as not even knowing what types of release can happen, and this is filled in by the server. Consider [`html/template`](https://pkg.go.dev/html/template) if necessary, but element updating could also be handled by js to keep it single-html/single-page.

cmd/releaseagent/main.go
```diff
@@ -15,14 +15,14 @@ import (
 )
 
 const description = `
-releaseagent coordinates a release of Microsoft build of Go repository.
+releaseagent coordinates a release of Microsoft Build of Go repository.
```
dagood:
```suggestion
releaseagent coordinates a release related to the Microsoft build of Go project.
```

cmd/releaseagent/internal/releasesteps/releasesteps.go
```diff
@@ -220,113 +470,204 @@ func CreateStepGraph(ri *Input, secret *Secret, rs *State, sb ServiceBundle) ([]
...
 			return fmt.Sprintf("%s, %s", n, version)
 		}
 
 		syncUpdate := coordinator.NewStep(
+			id("upstream-commit"),
 			name("⌚ Get upstream commit for release"),
 			noTimeout,
 			func(ctx context.Context) error {
-				if vs.UpstreamCommit != "" {
+				if stateValue(state, func(s *State) string { return s.Versions[version].UpstreamCommit }) != "" {
```
dagood: This seems like it complicates the way the step graph is put together, but it's not clear to me what the benefit is. Maybe the threading model has been changed? Admittedly it has been a bit, but I remember optimizing the design to not need explicit threading coordination like this.

dagood (commented):
A few initial comments, trying to focus on things that might make it harder to expand upon later.

cmd/releaseagent/internal/releaseui/web/go-images.html
michelle-clayton-work: I think this would be better addressed in the next iteration which will be focused on refactoring and making it easier to add new processes 

Still a draft but evenutally will be https://github.com/microsoft/go-infra/pull/536

cmd/releaseagent/internal/coordinator/step.go
michelle-clayton-work: removed -- the thought was that if someone changed the name for some reason then it would need to be changed in multiple places, but since we are the consumers and producers you're right that it doesn't really apply here

cmd/releaseagent/internal/coordinator/step.go
michelle-clayton-work: Yup removed

cmd/releaseagent/internal/releasesteps/releasesteps.go
michelle-clayton-work: Agreed -- restored direct state access in the legacy full-release graph and removed the checkpoint/mutex machinery from that path. The serialized state wrapper now remains only in the focused go-images workflow, where it supports durable checkpoints around queue intent, build ID, and completion.

cmd/releaseagent/internal/azdopipeline/client.go
michelle-clayton-work: moved the supported operations to the generated SDK. Azure Repos reads now use git.GetItem, git.GetCommit, and git.GetRefs; pipeline definition and timeline retrieval use build.GetDefinition and build.GetBuildTimeline.

The remaining custom REST is limited to build retrieval/listing and queueing because the generated build.Build model omits templateParameters. QueueBuild marshals that same model, so it cannot send them. It can send the legacy Parameters field used for correlation variables. releasego build-pipeline also uses a manual request for the template-parameter limitation.

cmd/releaseagent/internal/releasesteps/releasesteps.go
michelle-clayton-work: Added a focused GoImagesInput containing only the values needed by its flow. Input still remains for the existing flow

cmd/releaseagent/internal/coordinator/step.go
```diff
@@ -214,44 +110,67 @@ func (s *Step) TransitiveDependencies() ([]*Step, error) {
...
+			}
+			if _, ok := dependencies[dependency]; ok {
+				return fmt.Errorf("step %q depends on %q more than once", s.Name, dependency.Name)
+			}
+			dependencies[dependency] = struct{}{}
+			if err := visit(dependency); err != nil {
+				return err
 			}
 		}
-		// Mark current step as visited: we have ruled out cycles.
```
dagood: A lot of useful comments have been removed here. I think the extra error detection makes sense (and therefore introducing an explicit `path` stack) but the intent isn't always clear.

dagood:
We decided to go with a persistent dev branch here for now. With a large PR like this I feel like I'm being a little noisy and there's some friction in me being able to try out and suggest some kinds of larger-scale changes. This will make it easier to iterate before merging into a production-level branch (main).

And: we agreed this is a good starting point for that branch. 😄

michelle-clayton-work:
Closing this and tracking changes for relui on feature branch `dev/feature/relui`

PR #557, opened 2026-08-26
releaseagent: separate go-images workflow domain

michelle-clayton-work:
## Summary

Separate the standalone go-images release workflow and durable state from the broader Microsoft Go release model.

## Changes

- Add `internal/goimagesworkflow` with focused `Input`, `State`, DAG, pipeline parameters, and a three-method external service boundary.
- Move go-images execution and persistence onto the focused workflow model.
- Remove the standalone go-images DAG and fields from `releasesteps`.
- Persist only the current focused go-images session schema; remove prototype compatibility fields and migration code.
- Omit the informational `_info` pipeline parameter so Azure uses its default.
- Name the durable callback type `CheckpointFunc`.
- Group specialized go-images plan and session data under one `goImagesRuntime` field on `releaseui.Server`; generic processes such as go-infra continue to use `ProcessRun`.
- Reduce the disabled service adapter to the three operations required by go-images.

This removes the overlap between standalone go-images releases and the full multi-product release workflow. The release UI is not in active use, so this change intentionally does not preserve prototype session files.

## Validation

- golangci-lint v2.10.0
- `git diff --check`

No workflows, releases, labels, or Azure pipelines were triggered.

## Follow-ups

1. #558 - name the broader full-release workflow explicitly
2. #559 - trim registry and persistence plumbing
3. #563 - remove inferred ongoing-release tracking

Base: `dev/feature/relui`


cmd/releaseagent/serve.go
```diff
@@ -320,24 +320,24 @@ func processRunStorePath(sessionPath string) (string, error) {
 	return sessionPath + ".process-run.json", nil
 }
 
-func goImagesModeFromBuild(build *azdopipeline.Build) releasesteps.GoImagesReleaseMode {
+func goImagesModeFromBuild(build *azdopipeline.Build) goimagesworkflow.Mode {
```
dagood, line 323: I think this feature is something we don't need... it's the ability to take an Azure pipeline and attempt to reconstruct the release flow that includes it without any access to the true release configuration, right?

I think in some cases this works, but in the general case (additional release flows) this is very difficult (and perhaps not possible for complex or truly broken situations), and it seems like it is likely adding too much complexity and making it more difficult to split things up.

dagood: (Something to follow up with in another PR, I see this kind of change wouldn't fit in this PR.)

cmd/releaseagent/internal/goimagesworkflow/workflow.go
```diff
@@ -0,0 +1,299 @@
...
+const (
+	shortTimeout            = 10 * time.Minute
+	internalMirrorTimeout   = 16 * time.Minute
+	officialPipelineTimeout = 2 * time.Hour
+)
+
+// PipelineParameters derives the complete pipeline parameter set from an allowlisted mode.
+func PipelineParameters(mode Mode, sourceBuildID string) (map[string]string, error) {
+	parameters := map[string]string{
+		"_info":                    "🔵  go-docker-rolling-internal-pipeline.yml  🔵 🔵",
```
dagood: I think this should be omitted (left default), there should never be a reason to specify it.


```suggestion
```

cmd/releaseagent/internal/goimagesworkflow/workflow.go
```diff
@@ -0,0 +1,299 @@
...
+
+// Service is the complete external surface available to the standalone go-images workflow.
+type Service interface {
+	PollMirror(context.Context, string, string) error
+	QueuePipeline(context.Context, int, map[string]string) (string, error)
+	PollPipeline(context.Context, string) error
+}
+
+// Checkpoint durably records State. The state pointer is valid only during the call.
+type Checkpoint func(context.Context, *State) error
```
dagood: I'd call a func type like this `CheckpointFunc` to clarify that it's making a checkpoint, not a checkpoint itself. (If the word could only mean an action, I wouldn't insist on the suffix, but it confused me in context.)

cmd/releaseagent/internal/releasesteps/releasesteps.go
```diff
@@ -9,51 +9,31 @@ import (
...
-	"sync"
 	"time"
 
 	"github.com/microsoft/go-infra/cmd/releaseagent/internal/coordinator"
 )
 
 //go:generate go run github.com/matryer/moq -out ServiceBundle_moq_test.go . ServiceBundle
 
-// GoImagesReleaseMode identifies one explicitly allowlisted use of pipeline 1023.
+// GoImagesReleaseMode is retained only for schema-6 session compatibility.
```
dagood: Why is schema-6 session compatibility important? It seems counter to the PR goal. (Maybe you need to make it clear to Copilot that this tool isn't in active use so Copilot doesn't try to add backward compatibility?)

cmd/releaseagent/internal/releaseui/server.go
```diff
@@ -66,8 +67,8 @@ type Server struct {
 	input             PlanInput
 	source            GoImagesSource
 	rollbackSource    *GoImagesRollbackSource
-	releaseInput      *releasesteps.GoImagesInput
-	releaseState      *releasesteps.State
+	releaseInput      *goimagesworkflow.Input
+	releaseState      *goimagesworkflow.State
```
dagood: At first glance it's odd that a `releaseui.Server` type would have direct, individual references to the go images workflow input and state. I'm also wondering why it has go-images info but no go-infra info. Not sure what precisely is driving this, just an example of a design smell I'm looking out for.

dagood (commented):
Some comments from a somewhat-random read of the files. 😄

cmd/releaseagent/serve.go
michelle-clayton-work: https://github.com/microsoft/go-infra/pull/563

dagood (approved):

PR #570, opened 2026-08-29
releaseagent: narrow registry to active process contracts

michelle-clayton-work:
Stacked on #569.

Narrows the registry to the two release processes the UI can run today. Every registry entry must have a workflow, and the dashboard no longer carries unavailable-process status fields.

Browser inputs are limited to the contracts currently used by go-images and go-infra: fixed choices and positive integer IDs. The browser and server enforce the same positive-integer rule. The README now documents only the remaining command and supported contracts.

Validation:
- golangci-lint v2.10.0
- `git diff --check`

### DIFF
--- a/cmd/releaseagent/internal/releaseui/web/index.html
+++ b/cmd/releaseagent/internal/releaseui/web/index.html
@@ -35,10 +35,6 @@
           Release documentation <span aria-hidden="true">↗</span>
         </a>
       </div>
-      <div class="hero-stat" aria-label="Available release processes">
-        <strong id="available-count">1</strong>
-        <span>process available</span>
-      </div>
     </section>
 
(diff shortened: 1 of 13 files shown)


cmd/releaseagent/internal/releaseui/web/index.html
```diff
@@ -59,7 +55,7 @@ <h2 id="ongoing-title">Ongoing releases</h2>
         <div>
           <p class="eyebrow">Start new work</p>
           <h2 id="processes-title">Release processes</h2>
-          <p>Choose the product to release. More processes can be added without changing this dashboard.</p>
+          <p>Choose Go images or Go infrastructure.</p>
```
dagood: Regresses a bit on the process-agnostic core idea.

dagood (approved):

PR #578, opened 2026-09-02
releaseagent: narrow workflow metadata

michelle-clayton-work:
Stacked on #577.

## Summary

- require every registered process to provide preflight, plan, start, state, and event behavior
- remove capability flags that are constant for both current processes
- make every visible process input required
- retain simulation as the one optional workflow capability

## Why

Go images and go-infra both expose the complete release lifecycle. Modeling those routes and visible-input requirements as optional produced branches the current UI cannot use. Keeping only `CanSimulate` preserves the intentional go-images simulation without carrying hypothetical workflow shapes.

## Validation

- golangci-lint v2.10.0
- simulation route and UI coverage remain


michelle-clayton-work:
@copilot fix linter issues

dagood (approved):

PR #579, opened 2026-09-02
releaseagent: trim process response data

michelle-clayton-work:
Stacked on #578.

## Summary

- remove raw pipeline, source, rollback, timeout, timestamp, and result fields the embedded browser does not consume
- stop persisting pipeline parameters that are derived from immutable go-images input
- render go-images review data through the existing process-neutral view
- carry the process mark in dashboard summaries so go-infra runs display `IN`

## Why

The browser is the only consumer of these unreleased response contracts. Returning both raw domain objects and an already-normalized review view duplicated state and let tests depend on data the UI ignored. The retained response still includes the plan digest, session identity, rendered facts, step status and errors, and exact external-run links.

## Validation

- golangci-lint v2.10.0
- dashboard and both process pages smoke-tested


dagood (approved):

michelle-clayton-work:
@copilot resolve conflicts

PR #582, opened 2026-09-02
releaseagent: separate go-images server lifecycle

michelle-clayton-work:
Stacked on #581.

## Summary

- move go-images planning and review rendering to `go_images_plan.go`
- move go-images execution, persistence, and restart handling to `go_images_execution.go`
- leave shared routing, dashboard, simulation, SSE, authentication, and HTTP helpers in `server.go`

## Why

`server.go` mixed the shared HTTP server with the full specialized go-images lifecycle. This move-only split reduces that file from roughly 1,350 lines to about 720 and makes the two existing lifecycle styles easier to navigate.

No interfaces, control flow, responses, or tests change in this PR.

## Validation

- golangci-lint v2.10.0
- `git diff --check`


cmd/releaseagent/internal/releaseui/go_images_execution.go
```diff
@@ -0,0 +1,395 @@
...
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/microsoft/go-infra/cmd/releaseagent/internal/coordinator"
+	"github.com/microsoft/go-infra/cmd/releaseagent/internal/goimagessession"
+	"github.com/microsoft/go-infra/cmd/releaseagent/internal/goimagesworkflow"
+)
+
+func (s *Server) executionResponseLocked() executionResponse {
```
dagood: This method is defined in `go_images_execution.go`, but it's still a method on `Server`, and it's not clear to me how `executionResponseLocked` would be specific to the Go images build by its signature.

(Would some other workflow need its own `executionResponseLocked`? How would it add one?)

michelle-clayton-work: Good point. I renamed it to goImagesExecutionResponseLocked and renamed the adjacent plan helper for consistency. Only another specialized workflow would need its own helper; generic durable workflows use processRunResponseLocked

gdams (approved):

PR #585, opened 2026-09-02
Regenerate PR kind classifier with gh-aw v0.87.10

michelle-clayton-work:
Regenerate the reusable PR kind classifier with gh-aw v0.87.10.

The v0.83.4-generated lock hard-coded `claude-sonnet-4.6` as its fallback. That model disappeared from the Copilot provider catalog for the `agentic-workflows` integrator on September 1, causing classifier runs to fail with HTTP 400 before reading the PR. The newer compiler uses the provider-aware `auto` alias for agent execution and the `detection` alias for threat detection instead of pinning a concrete model.

The regeneration also updates the compiler-managed setup action to v0.87.10, Copilot CLI to 1.0.80, and firewall runtime to 0.28.10. The source workflow file is unchanged from `main`.

Validation:
- `gh aw compile classify-pull-request-kind`
- `gh aw compile --no-emit --validate classify-pull-request-kind`
- `git diff --check`

The compiler reports only the existing advisory that `workflow_dispatch` has no `concurrency.job-discriminator`.

After this merges, repositories pinned to the existing reusable-workflow SHA still need to update that pin.


.github/workflows/classify-pull-request-kind.md
```diff
@@ -1,6 +1,7 @@
 ---
 description: Classifies a pull request by every substantive change kind
 tracker-id: classify-pull-request-kind
+model: "gpt-5.4"
```
qmuntal: The classifier don't need to be particularly clever. Better don't specify any model.

qmuntal (changes requested):

qmuntal (approved):

PR #586, opened 2026-09-03
git-go-patch: embed an asInvoker Windows manifest

qmuntal:
## Motivation

Windows can apply installer-detection and Program Compatibility Assistant heuristics to executables that do not declare a requested execution level. This caused `git-go-patch.exe` to be misclassified as a failed installer and return error 740 before Go's `main` ran, even for help or version commands.

This change embeds a Windows application manifest with `requestedExecutionLevel="asInvoker"` and `uiAccess="false"`. It explicitly tells Windows to run with the caller's existing token, preventing heuristic elevation without requesting additional privileges.

## Implementation

The manifest is linked into `windows/386`, `windows/amd64`, and `windows/arm64` builds. The generated `.syso` files remain checked in so normal `go install` works without running a resource compiler.

A small standard-library-only generator lives in `internal/cmd/winmanifest`. We intentionally keep this in-repo rather than using an external module because `.syso` files are opaque native COFF linker inputs: owning the generator makes their construction auditable, avoids adding a build-time supply-chain dependency, and lets focused tests enforce that each object contains only the expected read-only manifest resource, local symbol, and self-relocation. `TestGeneratedFilesAreCurrent` byte-compares freshly generated output with the checked-in resources, and the normal PR `go test ./...` job enforces it.

## Validation

```text
go test ./internal/cmd/winmanifest -count=1
go test ./cmd/git-go-patch -run '^$' -count=1
CGO_ENABLED=0 cross-builds: windows/386, windows/amd64, windows/arm64, linux/amd64
go generate ./cmd/git-go-patch (pristine worktree)
```


.github/workflows/check-git-go-patch-generated.yml
```diff
@@ -0,0 +1,38 @@
...
+        shell: bash
+        run: |
+          set -euo pipefail
+          go generate ./cmd/git-go-patch
+          if [[ -n "$(git status --porcelain -- cmd/git-go-patch)" ]]; then
+            git status --short -- cmd/git-go-patch
+            git diff --stat -- cmd/git-go-patch
+            echo "::error::git-go-patch generated files are stale. Run 'go generate ./cmd/git-go-patch' and commit the results."
+            exit 1
+          fi
```
dagood, line 38: Can you run this as a test instead of CI-only?

dagood (changes requested):

internal/winmanifest/main.go
```diff
@@ -0,0 +1,314 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package main
```
dagood, line 4: Would `internal/cmd/winmanifest` make more sense for clarity?

.github/workflows/check-git-go-patch-generated.yml
qmuntal: Done. I removed the dedicated workflow and made freshness an ordinary Go test, TestGeneratedFilesAreCurrent. The normal PR go test ./... job now enforces the byte-for-byte generated output and COFF structure checks.

internal/cmd/winmanifest/main.go
```diff
@@ -0,0 +1,314 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package main
```
qmuntal: Done. I moved the generator and its tests to internal/cmd/winmanifest and updated the go:generate directive and documentation.

dagood (approved):

PR #590, opened 2026-09-03
Add compliance assessment completion tool

gdams:
## Summary

- add `cmd/compliance`, a CLI for recurring Azure DevOps compliance assessments
- reuse answers from the latest earlier completed assessment for each policy after validating them against the current questionnaire
- generate compliance work items, resume failed sessions, and complete only child activities that were completed in the selected source assessment
- support explicit, validated overrides when a questionnaire or activity set changes

## Use case

Recurring compliance assessment groups frequently recreate the same policy questionnaires and work-item activities. Completing them manually means finding the prior assessment, copying unchanged answers, recreating its work items, and closing activities that were already satisfied.

This command automates that repeatable work while treating the current assessment as authoritative. It selects prior assessments by policy ID, checks question IDs and definitions against the current graph, and refuses ambiguous changes. After saving answers, it generates a new completion session and reconciles child work items by activity node. Parent work items remain open, matching the portal's existing completion behavior.

## Safety and recovery

- dry-run is the default; `-apply` requires an explicit assessment-group name
- completed assessments are skipped, and in-progress assessments require `-overwrite`
- authentication comes from `az login` or `AZURE_DEVOPS_EXT_PAT`; credentials are not logged
- non-JSON error bodies are not echoed
- changed questionnaires require a complete `-answers-file` that is validated against all current questions and allowed option values
- current-only activities require an explicit `-complete-activity ASSESSMENT=NODE-ID` approval
- failed session generation is resumable and reruns are idempotent
- work items use the current Azure Boards project root instead of historical child paths that may have been deleted
- only node-matched or explicitly approved child work items are completed; parent work items are untouched

## Validation

- `go test ./...`
- `go test -race ./cmd/compliance`
- `golangci-lint run ./cmd/compliance/...`
- `go vet ./cmd/compliance`
- `go mod verify`
- end-to-end dry-run, apply, failure recovery, and idempotent rerun against an internal assessment group

### DIFF
--- a/cmd/compliance/README.md
+++ b/cmd/compliance/README.md
@@ -1,0 +1,39 @@
+# Compliance assessment completion tool
+
+`compliance` copies questionnaire answers from the most recent earlier completed assessment for each policy in an Azure DevOps compliance assessment group. It then saves each assessment, generates its work items, and completes child work items whose matching source activity was completed.
+
+The command uses the Azure DevOps API directly. Authentication comes from `az login`, or from `AZURE_DEVOPS_EXT_PAT` when using a PAT configured for the Azure DevOps CLI.
+
+Start with a read-only dry run:
+
+```sh
+go run ./cmd/compliance \
+  -url https://dev.azure.com/ORGANIZATION/PROJECT/_compliance/product/PRODUCT-ID/assessments
+```
+
+The dry run selects the newest assessment group, identifies the latest completed source for each policy, checks current question IDs, and validates the prior work-item configuration. It does not save assessments or create work items.
+
+After reviewing the output, pin the group and apply it:
+
+```sh
+go run ./cmd/compliance \
+  -url https://dev.azure.com/ORGANIZATION/PROJECT/_compliance/product/PRODUCT-ID/assessments \
+  -assessment-group ASSESSMENT-GROUP \
+  -apply
+```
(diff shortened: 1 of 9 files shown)


gdams:
@copilot address this feedback

dagood (commented):
IMO this belongs in an internal Microsoft repository, so we can easily add features and tweaks that are specific to internal aspects of the assessment process.

I don't think there's value in making this open source that would justify the extra maintenance cost of some kind of plugin system.

dagood:
See https://github.com/microsoft/go-lab/tree/main/devtool/cmd/workitemhyperlinklist for example.

gdams:
> IMO this belongs in an internal Microsoft repository, so we can easily add features and tweaks that are specific to internal aspects of the assessment process.
> 
> I don't think there's value in making this open source that would justify the extra maintenance cost of some kind of plugin system.

agreed, moving to go-lab

gdams:
Closing in favor of moving this command to microsoft/go-lab under devtool. The tool intentionally depends on internal Microsoft compliance APIs and workflow details, so go-lab is the better ownership and maintenance boundary. A replacement PR will follow there.

gdams:
Replacement PR: microsoft/go-lab#507

PR #592, opened 2026-09-08
relui: improve failure diagnostics

michelle-clayton-work:
Improves the diagnostics shown when release readiness checks or a go-images pipeline run fails.

## Summary

- replace the generic execution-unavailable message with details from non-passing preflight checks
- read the Azure timeline once after a failed go-images run and report up to three failed task paths with their issue messages
- retain the Azure run link and fall back to the run-level result when timeline details are unavailable

This does not restore continuous timeline polling or persist timeline data.

## Validation

- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.10.0 run ./cmd/releaseagent/...`
- `git diff --check`

### DIFF
--- a/cmd/releaseagent/internal/azdopipeline/client.go
+++ b/cmd/releaseagent/internal/azdopipeline/client.go
@@ -13,6 +13,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
@@ -60,6 +61,12 @@ type Build struct {
 	SourceVersion      string
 	Parameters         map[string]string
 	TemplateParameters map[string]any
+}
+
+// BuildFailure identifies one failed leaf in an Azure Pipelines run.
+type BuildFailure struct {
+	Path    string
+	Message string
 }
 
 // Definition is the allowlist-relevant metadata of an Azure Pipeline definition.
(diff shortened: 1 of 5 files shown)


michelle-clayton-work:
@dagood let me know if you think these changes are more beneficial or more clutter

cmd/releaseagent/internal/azdopipeline/client.go
```diff
@@ -335,6 +357,91 @@ type apiBuild struct {
 	Links json.RawMessage `json:"_links"`
 }
 
+type apiTimelineRecord struct {
+	ID       string `json:"id"`
```
dagood: This is a bit odd, does github.com/microsoft/azure-devops-go-api really not define this?

dagood (approved):
I think it's fine to surface the error list, we'll probably want to process things like this and show the runner some suggestions on how to proceed at some point.

cmd/releaseagent/internal/azdopipeline/client.go
michelle-clayton-work: You're right. Switched GetFailures to use those and removed the custom timeline DTO and raw timeline request. Raw REST remains for build reads because the SDK model omits templateParameters

PR #596, opened 2026-09-09
pipelineymlgen: preserve inline YAML map order

dagood:
Retain source YAML nodes for structured inline-template data so yml insertion does not reorder AzDO Pipeline keys.

For example, currently, alphanumeric ordering changes this:

```yml
      - stage: PrepareRelease
        jobs:
          - job: PrepareRelease
```

to:

```yml
      - jobs:
          - job: PrepareRelease
        stage: PrepareRelease
```

which AzDO doesn't accept.

Fixing this allows jobs to pass through `data`, improving the flexibility of `pipelineymlgen` templates.

---

Yes, AzDO pipeline yaml is not compliant with the YAML spec:
https://yaml.org/spec/1.2.2/#3221-mapping-key-order

> To serialize a mapping, it is necessary to impose an ordering on its keys. This order is a serialization detail and should not be used when composing the representation graph (and hence for the preservation of application data). **In every case where node order is significant, a sequence must be used.**

### DIFF
--- a/internal/pipelineymlgen/eval.go
+++ b/internal/pipelineymlgen/eval.go
@@ -33,6 +33,10 @@ type EvalState struct {
 	// inlinerange iterates without variable names, it may be any value
 	// (e.g. a scalar), allowing ${ . } to give the element directly.
 	Data any
+
+	// preservedYAML retains the original YAML nodes for maps and slices decoded
+	// into Data so yml can emit them without losing mapping order.
+	preservedYAML map[dataIdentity]preservedYAMLValue
 }
 
 // EvalFile is a helper to run EvalFileConfig and evalFileWithConfig in one.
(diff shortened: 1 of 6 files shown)


internal/pipelineymlgen/eval.go
```diff
@@ -33,6 +33,10 @@ type EvalState struct {
 	// inlinerange iterates without variable names, it may be any value
 	// (e.g. a scalar), allowing ${ . } to give the element directly.
 	Data any
+
+	// preservedYAML retains the original YAML nodes for maps and slices decoded
+	// into Data so yml can emit them without losing mapping order.
+	preservedYAML map[dataIdentity]preservedYAMLValue
```
dagood, line 39: I don't like this approach: very specific to one goal and seems fiddly/unflexible. https://github.com/microsoft/go-infra/pull/597 is a more natural fit, if it works out.

PR #599, opened 2026-09-10
releaseagent: persist generic runs in work items

michelle-clayton-work:
## Summary

- replace generic `ProcessRun` file persistence with explicit Azure DevOps work-item identity and revision handling
- create the work item only after confirmation and before the external GitHub action can mutate anything
- checkpoint and restore generic runs through revision-checked work-item updates
- assign new work items to the authenticated operator and classify confirmed go-infra dry runs as tests

This keeps unconfirmed plans in memory and preserves the existing preflight, confirmation, and target-reconciliation behavior.

## Stack

2 of 4. Stacked on #598.

## Testing

- `go test -count=1 ./cmd/releaseagent/...` at this layer
- `go test -count=1 -race ./cmd/releaseagent/...` on the completed stack
- `go vet ./cmd/releaseagent/...`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.10.0 run ./cmd/releaseagent/...`

PR #600, opened 2026-09-10
releaseagent: persist go-images runs in work items

michelle-clayton-work:
## Summary

- persist validated go-images session documents in the shared Azure DevOps work-item envelope
- create the work item before queueing pipeline 1023 and checkpoint queue intent, build discovery, and terminal state by expected revision
- restore known builds for monitoring without allowing a replacement queue request
- remove the go-images local session file, lease, atomic-file helper, and `-session-file` configuration

Test-mode executions remain real pipeline runs, but their work items carry immutable test classification and publication stays locked to `dev/`.

## Stack

3 of 4. Stacked on #599.

## Testing

- `go test -count=1 ./cmd/releaseagent/...` at this layer
- `go test -count=1 -race ./cmd/releaseagent/...` on the completed stack
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.10.0 run ./cmd/releaseagent/...`


michelle-clayton-work:
@copilot fix linter failures

michelle-clayton-work:
@copilot resolve merge conflicts

cmd/releaseagent/serve.go
```diff
@@ -100,20 +72,33 @@ func handleServe(parse subcmd.ParseFunc) error {
 	if err != nil {
 		return err
 	}
+	goImagesStore, err := releaseui.NewGoImagesWorkItemStore(workItems, assignedTo)
```
dagood, line 75: Removing special cases will help a lot with changes and review IMO.

dagood: Here's a new structure I'd try that I think might help crystallize a flexible approach:

* `/releaseui/...`
  * Implements the server, ui, and workitem state tracking.
  * Defines some `interface`s that a release process needs to implement in order to be tracked by the release UI.
* `/cmd/releaseui`
  * The thing that Microsoft build of Go devs "go run" to start releases and track them.
* `/cmd/releaseui/internal/goimagesrelease`
  * Implements the interfaces defined in `/releaseui/...` for an images-only release.
* `/cmd/releaseui/internal/goinfrarelease`
  * Implements the interfaces defined in `/releaseui/...` for a go-infra tag release.
* `/cmd/releaseui/internal/fullgorelease` (a next step, just to demonstrate the pattern)
  * Implements the interfaces defined in `/releaseui/...` for the full Microsoft build of Go release process.

Then `/cmd/releaseui` passes all these release types into `/releaseui/...` while calling the `/releaseui/...` API that hosts the server.

The `/cmd/releaseui/internal/*` packages are responsible for all the direct AzDO pipeline checks, GitHub Actions calls, etc., although they end up all using high level functionality from the `/azdo`, `/githubutil`, `/internal/...` etc. packages (with more functionality added there as needed). The release process packages do what they have to do to grab the info and shuffle it along through the interfaces defined in `/releaseui/...`.

This also means that other projects can define their own e.g. `github.com/microsoft/agent-framework-go/cmd/mafreleaseui` packages that define their own set of releases.

As a broader team we might also want to have a `github.com/microsoft/go-lab/cmd/allreleaseui` that includes every single type of release that goes through this process, as an all-up meta view. In that case we could make the release packages public rather than internal.

